PR #66, opened 2020-04-25
better rust github actions

ETeissonniere:
## Context
Improve the Github actions to be a bit more useful.


ETeissonniere:
Seems like whatever we are doing we always end up running out of space. Reached out to github for help. If someone has some tips let us know

PR #76, opened 2020-04-30
configure sccache for faster builds

ETeissonniere:
## Context
Configure `sccache` for faster rust builds.


ETeissonniere:
Seems like we have issues starting the server, while this would have been great I don't to spend too much time on this. Going back to github caches which are good enough and work from the ground up.

PR #293, opened 2020-09-29
upgrade node and runtime to 2.0

ETeissonniere:
## Context
Parity recently released Substrate 2.0.0. This PR bumps the node and runtime to this new version. We also add support for the `finality-tracker` pallet to detect finality stalls and maybe take some automated actions in the future. We also apply the new 2.0.0 configurations to the frame pallets in order to support a more fair and efficient weight computation system.

**TBD**:
- [x] check if we should update `parity-scale-codec`
- [x] check if we should update `serde`

### Sanity
- [x] I have incremented the runtime version number
- [ ] I have incremented `transaction_version` if needed

### Quality
- [ ] I have added unit tests, and they are passing
- [ ] I have added benchmarks to my pallet
- [ ] I have added the benchmarks to the node
- [x] I have added potential RPC calls to the node
- [x] I have eventual custom types to the `types.json` file
- [ ] I have added comments and documentation

### Testing
- [ ] The node runs fine on a development network
- [ ] The node runs fine on a local network
- [ ] The node runs fine on an upgraded local network
- [x] The node can synchronize the existing network

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2308,47 +2308,18 @@ dependencies = [
 
 [[package]]
 name = "jsonrpc-client-transports"
-version = "14.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2773fa94a2a1fd51efb89a8f45b8861023dbb415d18d3c9235ae9388d780f9ec"
-dependencies = [
- "failure",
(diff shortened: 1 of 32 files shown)


ETeissonniere:
Still need to test syncing etc before pushing to arcadia.

PR #299, opened 2020-10-20
fix mainnet stall

ETeissonniere:
## Context
Our validators stalled for too long and this broke the babe consensus.
We used a patched version of `fork-off-substrate` to duplicate the chain's state and create this new chainspec. Once the release is available as a docker container we will update and restart the validators.


ETeissonniere:
CI failure is due to the Github Cache, not our test. Forcing the merge.

PR #321, opened 2021-04-05
pallet migration to FRAME v2 & refactor weights

### LINKED ISSUE #102: Refactor Weights
## Describe the enhancement
After #88 refactor the pallet's weights and make sure they make sense

shamb0:
pallet migration to FRAME v2 & refactor weights

fixes #102


pallets/amendments/src/lib.rs
```diff
@@ -21,88 +21,79 @@
...
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
         /// Schedule `amendment` to be executed after the configured time, unless vetoed by `VetoOrigin`
-        #[weight = 100_000_000]
-        fn propose(origin, amendment: Box<T::Amendment>) -> DispatchResult {
+        #[pallet::weight(T::WeightInfo::propose(*_length_bound))]
+        pub fn propose(
+            origin: OriginFor<T>,
+            amendment: Box<T::Amendment>,
+            _length_bound: u32,
```
ETeissonniere, line 88: I think we should check that `length_bound >= size_of_amendment` and throw an error if this is not the case

ETeissonniere (changes requested):
Thanks! I only have one tiny comment :)

pallets/amendments/src/lib.rs
shamb0: Super Fast, Thanks for quick comment. Sure will rework on it :+1:

shamb0: Hello Eliott,

Completed refactoring weights annotation for `pub fn propose()`, as discussed removed the extra argument of ` _length_bound: u32`.

ETeissonniere (approved):
Very nice, thanks!

PR #328, opened 2021-04-21
Patches to Merge parachain and main branch

### LINKED ISSUE #308: Merge parachain and main branch
For now we have an `eliott/parachain` and a `master` branch. We may want to look into merging both of them together and have a set of command flags to either start a parachain or a standalone chain to avoid rebasing issues.

rajesh-nodle:
Fixes #308


runtime/src/lib.rs
```diff
@@ -964,6 +300,7 @@ sp_api::impl_runtime_apis! {
             add_benchmark!(params, batches, pallet_amendments, Amendments);
             add_benchmark!(params, batches, pallet_balances, Balances);
             add_benchmark!(params, batches, pallet_collective, TechnicalCommittee);
+            add_benchmark!(params, batches, pallet_contracts, Contracts);
```
ETeissonniere, line 303: This wouldn't work if the `Contracts` pallet is not in the runtime

runtime/src/lib.rs
```diff
@@ -24,733 +24,40 @@
...
-    type NativeCurrency = NodleToken;
-    type GetNativeCurrencyId = GetNodleTokenId;
-    type WeightInfo = ();
-}
-
-impl orml_unknown_tokens::Config for Runtime {
-    type Event = Event;
-}
-
 construct_runtime!(
```
ETeissonniere, line 61: Are we not merging the changes with the custom macros?

node/src/rpc.rs
```diff
@@ -47,11 +47,13 @@ where
     C: HeaderBackend<Block> + HeaderMetadata<Block, Error = BlockChainError> + 'static,
     C: Send + Sync + 'static,
     C::Api: substrate_frame_rpc_system::AccountNonceApi<Block, AccountId, Index>,
+    // C::Api: pallet_contracts_rpc::ContractsRuntimeApi<Block, AccountId, Balance, BlockNumber>,
```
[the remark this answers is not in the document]
rajesh-nodle: No Eliott, build is OK with this comment expression

runtime/src/lib.rs
rajesh-nodle: Sorry overlooked, correction done :+1:

runtime/src/lib.rs
rajesh-nodle: You are right, right now it's more like stubbed or disabled with partial implementation of '#[cfg(feature = "enable-contracts")]', Idea is to remove the annotation once package dependency is resolved.

And we can't use CLI args like `--features enable-contracts` to control the build.

PR #347, opened 2021-06-22
Preserve the state of `AwardedPts` for bonded session duration ...

rajesh-nodle:
`pay_stakers()`  is modified, such that it uses `iter()` is used instead of `drain()` to maintain the state of `AwardedPts` for bonded session duration.


ETeissonniere (changes requested):
can we unit test this to avoid a regression?

rajesh-nodle:
> can we unit test this to avoid a regression?

:+1:  Done Eliott

ETeissonniere (approved):
lgtm

PR #354, opened 2021-06-29
phase-0 chaos test runtime migration updates

rajesh-nodle:
phase-0 chaos test runtime migration updates

### DIFF
--- a/pallets/nodle-staking/src/migrations.rs
+++ b/pallets/nodle-staking/src/migrations.rs
@@ -19,14 +19,11 @@ use super::{BalanceOf, Config};
 use crate::types::Validator;
 use frame_support::{
     pallet_prelude::*,
-    traits::{Currency, Get, StorageInstance},
+    traits::{Get, StorageInstance},
     weights::{constants::RocksDbWeight, Weight},
 };
 
-use sp_runtime::{
-    traits::{AccountIdConversion, Zero},
-    Perbill,
-};
+use sp_runtime::{traits::Zero, Perbill};
 
 use sp_std::vec::Vec;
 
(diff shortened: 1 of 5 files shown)


pallets/nodle-staking/src/migrations.rs
```diff
@@ -113,8 +153,22 @@ pub fn poa_validators_migration<T: Config>() -> Weight {
...
             } else {
-                log::error!("Error!!! Staking Invulnerable invalid DB instance state");
-                return weight;
+                log::info!("Staking runtime upgrade Genesis config");
+
+                <StakingInvulnerables<T>>::put(poa_validators.clone());
+
+                // Ensure balance is >= ED
+                let imbalance = T::Currency::issue(T::Currency::minimum_balance());
+                T::Currency::resolve_creating(&T::PalletId::get().into_account(), imbalance);
```
ETeissonniere, line 162: I would like to avoid creating coins if possible due to the fact we are supposed to have a fixed supply. I would suggest checking wether we really need to do that, if we really need to let's take the funds from the company reserve instead

runtime/src/migrations.rs
```diff
@@ -83,3 +86,22 @@ impl OnRuntimeUpgrade for GrantsMigration {
...
+        debug::RuntimeLogger::init();
+        debug::print!("🕊️ Starting staking migration...");
+
+        weight = weight.saturating_add(NodleStaking::on_runtime_upgrade());
+
+        debug::print!("🕊️ Staking migration Done");
+
+        weight
+    }
+}
```
ETeissonniere, line 107: I don't think we need to call the on_runtime_upgrade function? Should be called by substrate already

ETeissonniere (changes requested):

pallets/nodle-staking/src/migrations.rs
rajesh-nodle: Indeed not required :+1:

runtime/src/migrations.rs
rajesh-nodle: Nice one, removed the messy stuff  Done :+1:

ETeissonniere (approved):

PR #370, opened 2021-07-09
New Api | Allow nominator to switch nomination between validators

rajesh-nodle:
New Api | Allow nominator to switch nomination between validators


ETeissonniere (changes requested):
Let's add some unit tests for it

PR #374, opened 2021-07-14
AlienAnalyticTokoin connect to earn right away.

twins2020:


ETeissonniere:
Hey, are you sure this is what you wanted to submit?
I see some binary files, a Covid vaccination certificate, and the terraform workflow won't work as-is since we have no terraform files in this repo.

I will close this PR for now, feel free to reopen or submit a new one if I was mistaken.

PR #375, opened 2021-07-16
Stake dynamic threshold check functionality

rajesh-nodle:
Stake dynamic threshold check functionality


pallets/nodle-staking/src/lib.rs
```diff
@@ -193,6 +195,52 @@ pub mod pallet {
...
+        #[pallet::weight(T::WeightInfo::set_staking_limits())]
+        pub fn set_staking_limits(
+            origin: OriginFor<T>,
+            max_stake_validators: u32,
+            min_stake_session_selection: BalanceOf<T>,
+            min_validator_bond: BalanceOf<T>,
+            min_nominator_total_bond: BalanceOf<T>,
+            min_nominator_chill_threshold: BalanceOf<T>,
+        ) -> DispatchResultWithPostInfo {
+            T::CancelOrigin::try_origin(origin)
```
ETeissonniere, line 207: Why `CancelOrigin`?

pallets/nodle-staking/src/lib.rs
```diff
@@ -193,6 +195,52 @@ pub mod pallet {
...
+                old,
+                min_nominator_total_bond,
+            ));
+
+            let old = <StakingMinNominationChillThreshold<T>>::get();
+            <StakingMinNominationChillThreshold<T>>::set(min_nominator_chill_threshold);
+            Self::deposit_event(Event::StakingMinNominationChillThreshold(
+                old,
+                min_nominator_chill_threshold,
+            ));
```
ETeissonniere, line 238: We can probably have one big event for the all the function instead of triggering lots of them

pallets/nodle-staking/src/lib.rs
```diff
@@ -388,6 +460,7 @@ pub mod pallet {
             origin: OriginFor<T>,
             validator: T::AccountId,
             amount: BalanceOf<T>,
+            unfreeze_bond: bool,
```
ETeissonniere, line 463: I am not clear what this parameter is?

pallets/nodle-staking/src/lib.rs
rajesh-nodle: Good question Eliott, 
For staking config api's like 
`set_invulnerables() | set_total_validator_per_round() | slash_cancel_deferred()` 
we made design choice of origin could be either `Root or CancelOrigin`.
Here `set_staking_limits()` also falls under same class, so caller origin 
could be either `Root or CancelOrigin`.

pallets/nodle-staking/src/lib.rs
rajesh-nodle: Done :+1:

pallets/nodle-staking/src/lib.rs
rajesh-nodle: As discussed

* Regarding argument unfreeze_bond in API

```
pub fn nominator_nominate(
    origin: OriginFor<T>,
    validator: T::AccountId,
    amount: BalanceOf<T>,
    unfreeze_bond: bool,
) -> DispatchResultWithPostInfo;

pub fn nominator_bond_more(
    origin: OriginFor<T>,
    validator: T::AccountId,
    more: BalanceOf<T>,
    unfreeze_bond: bool
) -> DispatchResultWithPostInfo;
```
As part of reconciliation, if nomination bond value is below threshold



nomination is removed from validator pool
and dust or residual value is moved to new variable frozen_bond ...



This value can be unlocked back as part of
nominator_nominate() & nominator_bond_more() using the new arg flag unfreeze_bond: bool.

And as suggested introduced new api `unbond_frozen()`, to with draw the amount.

pallets/nodle-staking/src/lib.rs
ETeissonniere:
```suggestion
            ensure_root(origin)?;
```

ETeissonniere: let's just `ensure_root` then

ETeissonniere (approved):

PR #412, opened 2021-10-06
pallet::event type or metadata changes to adopt for chain-minions client

rajesh-nodle:
Types or metadata part of pallet::event are modified to adopt
"substrate-subext" EventTypeRegistry

### DIFF
--- a/pallets/grants/src/lib.rs
+++ b/pallets/grants/src/lib.rs
@@ -49,6 +49,7 @@ pub type BalanceOf<T> =
     <<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
 pub type VestingScheduleOf<T> =
     VestingSchedule<<T as frame_system::Config>::BlockNumber, BalanceOf<T>>;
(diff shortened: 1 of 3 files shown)


pallets/grants/src/lib.rs
```diff
@@ -49,6 +49,7 @@ pub type BalanceOf<T> =
     <<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
 pub type VestingScheduleOf<T> =
     VestingSchedule<<T as frame_system::Config>::BlockNumber, BalanceOf<T>>;
+pub type VestingScheduleOfColl<T> = Vec<VestingScheduleOf<T>>;
```
ETeissonniere, line 52: If possible, I think we should rename to something that is more meaningful. Maybe `VestingSchedulesOf<T>` or `ListVestingScheduleOf<T>` for instance?

rajesh-nodle: Sure Eliott will do that :+1:

ETeissonniere (approved):

PR #427, opened 2021-11-15
runtime main & staking integration to node

rajesh-nodle:
Runtime main & staking integration to node


node/src/cli.rs
```diff
@@ -16,9 +16,70 @@
...
+
+    pub fn is_staking_runtime(&self) -> bool {
+        match self {
+            Self::Main => false,
+            Self::Staking => true,
+        }
+    }
+}
+
+impl fmt::Display for RuntimeInstance {
```
noisefloors, line 52: This trait should be accessible via `#[derive(fmt::Display)]` unless the manual implementation is preferred.

node/src/chain_spec/cs_staking.rs
```diff
@@ -0,0 +1,224 @@
...
+#[cfg(test)]
+pub(crate) mod tests {
+    use super::*;
+    use sp_runtime::BuildStorage;
+
+    #[test]
+    fn test_create_staking_local_chain_spec() {
+        local_staking_config().build_storage().unwrap();
+    }
+}
```
ETeissonniere, line 224: we should test the dev config if it is used too

node/src/command.rs
```diff
@@ -52,20 +51,29 @@ impl SubstrateCli for Cli {
...
+            Ok(match id {
+                "dev" => Box::new(chain_spec::cs_main::development_config()),
+                "local" => Box::new(chain_spec::cs_main::local_testnet_config()),
+                "" | "main" => Box::new(chain_spec::cs_main::main_config()),
+                "arcadia" => Box::new(chain_spec::cs_main::arcadia_config()),
+                path => Box::new(chain_spec::cs_staking::ChainSpec::from_json_file(
+                    std::path::PathBuf::from(path),
+                )?),
+            })
+        }
```
ETeissonniere, line 72: I think we should have a more standard setup where we define the following chain specs, without the runtime cli arg:
- `staking-dev`
- `staking-local`

node/src/command.rs
```diff
@@ -52,20 +51,29 @@ impl SubstrateCli for Cli {
...
     }
 
     fn native_runtime_version(_: &Box<dyn ChainSpec>) -> &'static RuntimeVersion {
-        &runtime_main::VERSION
+        &main_runtime::VERSION
     }
```
ETeissonniere, line 77: depends on the runtime right>

primitives/src/lib.rs
```diff
@@ -64,6 +64,8 @@ pub type Header = generic::Header<BlockNumber, BlakeTwo256>;
 pub type Block = generic::Block<Header, OpaqueExtrinsic>;
 /// Block ID.
 pub type BlockId = generic::BlockId<Block>;
+/// Type for storing the id of an asset.
+pub type AssetId = u32;
```
ETeissonniere, line 68: we do not have the asset pallet right?

runtimes/main/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Eliott Teissonniere <git.eliott@teissonniere.org>"]
 edition = "2018"
-name = "runtime-main"
+name = "main-runtime"
```
ETeissonniere, line 4: why change it? :)

runtimes/staking/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Eliott Teissonniere <git.eliott@teissonniere.org>"]
 edition = "2018"
-name = "runtime-staking"
+name = "staking-runtime"
```
ETeissonniere, line 4: why change it? :)

ETeissonniere (changes requested):

node/src/cli.rs
```diff
@@ -16,9 +16,70 @@
...
+
+#[derive(Debug, StructOpt)]
+pub enum RuntimeInstance {
+    Main,
+    Staking,
+}
+
+impl RuntimeInstance {
+    fn variants() -> [&'static str; 2] {
+        ["main", "staking"]
```
aliXsed, line 41: As a suggestion, you can add "derive(EnumIter)" from the strum package to RuntimeInstance and return a vector here by collecting all the values of the enum.

node/src/cli.rs
```diff
@@ -16,9 +16,70 @@
...
+    }
+}
+
+impl Default for RuntimeInstance {
+    fn default() -> Self {
+        RuntimeInstance::Main
+    }
+}
+
+impl FromStr for RuntimeInstance {
```
aliXsed, line 67: strum would also provide all the conversions to and from String for your enum. So you can remove this code if you use that.

node/src/cli.rs
```diff
@@ -30,6 +91,16 @@ pub struct Cli {
     pub run: RunCmd,
 }
 
+#[derive(Debug, StructOpt)]
+pub struct RunCmd {
+    #[structopt(flatten)]
+    pub base: sc_cli::RunCmd,
```
aliXsed, line 97: Smart! nice!

node/src/client.rs
```diff
@@ -0,0 +1,396 @@
...
+        + sp_block_builder::BlockBuilder<Block>
+        + frame_system_rpc_runtime_api::AccountNonceApi<Block, AccountId, Index>
+        + pallet_transaction_payment_rpc_runtime_api::TransactionPaymentApi<Block, Balance>
+        + sp_api::Metadata<Block>
+        + sp_offchain::OffchainWorkerApi<Block>
+        + sp_session::SessionKeys<Block>
+        + sp_authority_discovery::AuthorityDiscoveryApi<Block>,
+    <Self as sp_api::ApiExt<Block>>::StateBackend: sp_api::StateBackend<BlakeTwo256>,
+{
+}
```
aliXsed, line 55: Looks like we have some redundancy here. I read it like: implement all the traits listed above for `Api` if `Api` is already implementing them!

node/src/client.rs
```diff
@@ -0,0 +1,396 @@
...
+        + Sync
+        + CallApiAt<Block, StateBackend = Backend::State>,
+    Client::Api: RuntimeApiCollection<StateBackend = Backend::State>,
+{
+}
+
+/// Execute something with the client instance.
+///
+/// As there exist multiple chains inside MainRT, like MainRT itself and testing runtime,
+/// there can exist different kinds of client types. As these client types differ in the generics
```
aliXsed, line 95: Could we instead use the same client with all these different runtimes? Any api that the client should implement to match with a runtime, we can make sure the single client would implement all of them. It would then reduce the complexity of creating an matching two different clients for two different runtimes here.

node/src/client.rs
noisefloors: If I'm understanding correctly, this is a marker trait. As a suggestion, I think it makes sense to rename the generic parameter to something a bit clearer. `A` would be acceptable. If we're implementing generic runtime requirements, then maybe `Runtime` is appropriate?

node/src/cli.rs
noisefloors: ☝🏻 I'd forgotten about [`strum`](https://docs.rs/strum/0.22.0/strum/). I second the recommendation of choosing a library to simplify manual implementations.

node/Cargo.toml
```diff
@@ -3,67 +3,82 @@ authors = ["Eliott Teissonniere <git.eliott@teissonniere.org>"]
 build = "build.rs"
 edition = "2018"
 name = "nodle-chain"
-version = "2.0.13"
+version = "2.0.12"
+
+[package.metadata.wasm-pack.profile.release]
+# `wasm-opt` has some problems on linux, see
+# https://github.com/rustwasm/wasm-pack/issues/781 etc.
```
noisefloors, line 10: I appreciate this note about the process and reasoning for setting this option. 💯

node/src/client.rs
```diff
@@ -0,0 +1,396 @@
...
+        match self {
+            Self::MainRT(client) => T::execute_with_client::<_, _, FullBackend>(t, client.clone()),
+            Self::StakingRT(client) => {
+                T::execute_with_client::<_, _, FullBackend>(t, client.clone())
+            }
+        }
+    }
+}
+
+impl sc_client_api::UsageProvider<Block> for Client {
```
aliXsed, line 150: If we only take the FullClient out of a Client entry and pass it to the places where they need a client, it would then already have these traits implemented for. So we should be able to avoid implementing the traits like this for the enum. What do you think @rajesh-nodle?

rajesh-nodle:
Wow, great, thanks for quick review :+1:

aliXsed:
> Wow, great, thanks for quick review 👍

@rajesh-nodle I appreciate that developing inside a code like here where we inherit a lot of constraints from the definition and design of the upstream types and objects is very challenging. So please feel free to look at my comments as just suggestions and if you find it hard to change a working solution, it's fine to me. Nice job by the way 👍

node/src/command.rs
rajesh-nodle: The reason behind introduction of `--runtime` cli, is to handle the case of loading the chainspec from file path. The type `ChainSpec` is runtime specific, & this extra option is used to distinguish between the runtime to choose.

node/src/cli.rs
ETeissonniere: as long as it supports `no_std` that's a great idea!

rajesh-nodle: As suggested, extra cli options removed

node/src/cli.rs
rajesh-nodle: As suggested extra CLI options removed :+1:

node/src/command.rs
rajesh-nodle: Code refactor done, chain is identified from chain spec ID. introduced trait `IdentifyChain` to handle it :+1:

primitives/src/lib.rs
rajesh-nodle: Its my Bad, removed it :+1:

node/src/chain_spec/cs_staking.rs
rajesh-nodle: My Bad, added now. Done :+1:

node/src/cli.rs
rajesh-nodle: As suggested, extra CLI options removed :+1:

node/src/command.rs
rajesh-nodle: Good one, restored back.
Indeed its my bad, while implementing below chain identification api,
I was using `starts_with()`, so changed the name of crate.

```
fn is_runtime_staking(&self) -> bool {
	self.id().to_lowercase().starts_with("staking")
}
```

But finally id is picked from chain spec definition ...

```
ChainSpec::from_genesis(
	"Staking Local Testnet",
	"staking_local_testnet",
	ChainType::Local,
	local_staking_genesis,
	vec![],
	None,
	Some("nodl"),
	Some(build_local_properties()),
	Default::default(),
)
```

runtimes/staking/Cargo.toml
rajesh-nodle: Good one, restored back.
Indeed its my bad, while implementing below chain identification api,
I was using `starts_with()`, so changed the name of crate.

```
fn is_runtime_staking(&self) -> bool {
	self.id().to_lowercase().starts_with("staking")
}
```

But finally id is picked from chain spec definition ...

```
ChainSpec::from_genesis(
	"Staking Local Testnet",
	"staking_local_testnet",
	ChainType::Local,
	local_staking_genesis,
	vec![],
	None,
	Some("nodl"),
	Some(build_local_properties()),
	Default::default(),
)
```

runtimes/main/Cargo.toml
rajesh-nodle: Good one, restored back.
Indeed its my bad, while implementing below chain identification api,
I was using `starts_with()`, so changed the name of crate.

```
fn is_runtime_staking(&self) -> bool {
	self.id().to_lowercase().starts_with("staking")
}
```

But finally id is picked from chain spec definition ...

```
ChainSpec::from_genesis(
	"Staking Local Testnet",
	"staking_local_testnet",
	ChainType::Local,
	local_staking_genesis,
	vec![],
	None,
	Some("nodl"),
	Some(build_local_properties()),
	Default::default(),
)
```

node/src/client.rs
rajesh-nodle: Good Question, :-). Right now client is adopted to upstream code.

node/src/client.rs
rajesh-nodle: Good Question, :-). Right now client is adopted to upstream code.

node/Cargo.toml
```diff
@@ -3,67 +3,82 @@ authors = ["Eliott Teissonniere <git.eliott@teissonniere.org>"]
 build = "build.rs"
 edition = "2018"
 name = "nodle-chain"
-version = "2.0.13"
+version = "2.0.12"
```
ETeissonniere, line 6: why are we lowering the version?

ETeissonniere (approved):

node/Cargo.toml
rajesh-nodle: Thanks, @ETeissonniere, it's my bad. Restored it back :+1:

rajesh-nodle:
Observed event messages are displayed proper on polkadot JS app master with commit,  

fe215bc25a989189ddbe46f2b6ccdcc8690c0a4f

image

image

ETeissonniere (approved):

PR #439, opened 2021-12-02
Addition of chain spec "eden & eden-valid"

rajesh-nodle:
* Patch for error in 'build-spec' subcommands.
* Introduced "eden-util" chainspec as to generate Live chainspec for eden.

In the process creating production script, 
Got stuck in generating session key for `Aura`. 
so introduced this utility for generating the template json.


node/res/eden.json
```diff
@@ -0,0 +1,88 @@
+{
+  "name": "ParaChain Local Development",
```
ETeissonniere, line 2: can we use / have the non raw spec for review?

node/src/chain_spec/cs_eden.rs
```diff
@@ -214,6 +197,66 @@ pub fn local_config(id: ParaId) -> ChainSpec {
...
+    let valid_01_seed =
+        "castle fiscal easy grief horse include alley next youth deny engine response";
+    let valid_02_seed =
+        "giggle august whisper someone zoo trend evolve response bird board gaze pull";
+    let valid_03_seed =
+        "ramp clever visit pottery thunder fatal crawl already syrup gaze pride cabin";
+    let valid_04_seed =
+        "budget shock sudden axis predict soldier elder valve fame theme deliver slow";
+    let valid_05_seed = "party history head few eye bird exit axis flavor coconut kick robust";
+
```
ETeissonniere, line 210: we probably do not want mnemonics to be publicly available here :)

node/src/chain_spec/cs_eden.rs
```diff
@@ -214,6 +197,66 @@ pub fn local_config(id: ParaId) -> ChainSpec {
...
+        )]),
+        id.into(),
+    )
+}
+
+/// Local config, as close as possible to production (5 collators, no balances)
+pub fn valid_config(id: ParaId) -> ChainSpec {
+    ChainSpec::from_genesis(
+        "ParaChain Local Development",
+        "para_eden_local",
```
ETeissonniere, line 246: We should update the names

node/src/chain_spec/cs_eden.rs
```diff
@@ -214,6 +197,66 @@ pub fn local_config(id: ParaId) -> ChainSpec {
...
+        ChainType::Local,
```
ETeissonniere, line 247: should be `Live` no?

node/src/command.rs
```diff
@@ -83,14 +83,18 @@ fn load_spec(
...
         "dev" => Box::new(chain_spec::cs_main::development_config()),
         "local" => Box::new(chain_spec::cs_main::local_testnet_config()),
         "staking-dev" => Box::new(chain_spec::cs_staking::development_config()),
         "staking-local" => Box::new(chain_spec::cs_staking::local_staking_config()),
+        "eden" => Box::new(chain_spec::cs_eden::ChainSpec::from_json_bytes(
+            &include_bytes!("../res/eden.json")[..],
+        )?),
         "eden-dev" => Box::new(chain_spec::cs_eden::development_config(para_id)),
         "eden-local" => Box::new(chain_spec::cs_eden::local_config(para_id)),
+        "eden-valid" => Box::new(chain_spec::cs_eden::valid_config(para_id)),
```
ETeissonniere, line 97: I am not clear what is the purpose of `eden-valid`?

node/src/chain_spec/cs_eden.rs
noisefloors: I agree, I think these values should be provided as environment secrets and injected into the build if necessary.

node/src/command.rs
aliXsed: I guess it's the spec for westend. In that case eden-westend could be a better name. My question is what is the use case of eden-local vs eden-dev?

node/src/chain_spec/cs_eden.rs
rajesh-nodle: HaHa, thought of having common validation keys  for team, allocated only for internal validation, so that like alice & bob in polkadot js , it's available once injected. If it doesn't look good. will kick it out.

node/src/chain_spec/cs_eden.rs
rajesh-nodle: Good one, changed to

```
"name": "ParaChain Eden Live",
"id": "para_eden_live",
```

node/src/chain_spec/cs_eden.rs
rajesh-nodle: True, it should be `ChainType::Live`

node/src/command.rs
rajesh-nodle: Changed to "eden-util", for the purpose mentioned 
in description of the PR. I hope I am not over killing the code.

noisefloors: If we plan on removing it at some point in the future, then this placeholder seems acceptable.

node/src/chain_spec/cs_eden.rs
noisefloors: You should be able to use the [`env!()` macro](https://doc.rust-lang.org/std/macro.env.html) in cases where panicking on an unprovided value is acceptable.

ETeissonniere:
@rajesh-nodle I am going to go ahead and close this in favor of #442. The patch was already backported to `master`, and have already a simpler spec.

PR #444, opened 2021-12-12
feat: support try-runtime for all our runtimes

aliXsed:
Ready for your review after passing the following tests:

- [x] Test on-runtime-upgrade sub command with parachain
- [x] Test execute-block sub command with parachain
- [x] Test on-runtime-upgrade with mainnet
- [x] Test try-runtime as a CI job on each PR

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -87,7 +87,7 @@ pub enum Subcommand {
     #[structopt(name = "benchmark", about = "Benchmark runtime pallets.")]
     Benchmark(frame_benchmarking_cli::BenchmarkCmd),
 
-    /// Try some command against runtime state.
+    /// Try some command against runtime state. Note: `try-runtime` feature must be enabled.
     #[cfg(feature = "try-runtime")]
     TryRuntime(try_runtime_cli::TryRuntimeCmd),
 
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -87,6 +87,14 @@ pub enum Subcommand {
     #[structopt(name = "benchmark", about = "Benchmark runtime pallets.")]
     Benchmark(frame_benchmarking_cli::BenchmarkCmd),
 
+    /// Try some command against runtime state.
+    #[cfg(feature = "try-runtime")]
+    TryRuntime(try_runtime_cli::TryRuntimeCmd),
+
+    /// Try some command against runtime state. Note: `try-runtime` feature must be enabled.
+    #[cfg(not(feature = "try-runtime"))]
+    TryRuntime,
+
     /// Verify a signature for a message, provided on STDIN, with a given (public or secret) key.
     Verify(VerifyCmd),
 
--- a/node/src/command.rs
+++ b/node/src/command.rs
@@ -16,7 +16,17 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#[cfg(feature = "try-runtime")]
+use crate::para_service::EdenParachainRuntimeExecutor;
+#[cfg(feature = "try-runtime")]
+use crate::service::StakingExecutorDispatch;
+#[cfg(feature = "try-runtime")]
+use sc_executor::NativeExecutionDispatch;
+#[cfg(feature = "try-runtime")]
+use try_runtime_cli::TryRuntimeCmd;
+
 use crate::cli_relaychain::RelayChainCli;
+
 use crate::service::MainExecutorDispatch;
 use crate::{
     chain_spec,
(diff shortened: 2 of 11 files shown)


runtimes/staking/src/lib.rs
```diff
@@ -389,6 +392,21 @@ sp_api::impl_runtime_apis! {
             Ok(batches)
         }
     }
+
+    #[cfg(feature = "try-runtime")]
+    impl frame_try_runtime::TryRuntime<Block> for Runtime {
+        fn on_runtime_upgrade() -> (Weight, Weight) {
+            // NOTE: intentional unwrap: we don't want to propagate the error backwards, and want to
+            // have a backtrace here. If any of the pre/post migration checks fail, we shall stop
+            // right here and right now.
```
noisefloors, line 401: Great explanation. 💯

aliXsed: The credit goes to substrate :)

node/src/cli.rs
```diff
@@ -87,6 +87,14 @@ pub enum Subcommand {
     #[structopt(name = "benchmark", about = "Benchmark runtime pallets.")]
     Benchmark(frame_benchmarking_cli::BenchmarkCmd),
 
+    /// Try some command against runtime state.
+    #[cfg(feature = "try-runtime")]
+    TryRuntime(try_runtime_cli::TryRuntimeCmd),
+
+    /// Try some command against runtime state. Note: `try-runtime` feature must be enabled.
+    #[cfg(not(feature = "try-runtime"))]
+    TryRuntime,
```
ETeissonniere, line 96: Why the two commands?

node/src/command.rs
```diff
@@ -436,6 +446,37 @@ pub fn run() -> Result<()> {
...
+            } else if runner.config().chain_spec.is_runtime_staking() {
+                async_exec_try_runtime::<StakingExecutorDispatch>(runner, cmd)
+            } else {
+                async_exec_try_runtime::<MainExecutorDispatch>(runner, cmd)
+            }
+        }
+        #[cfg(not(feature = "try-runtime"))]
+        Some(Subcommand::TryRuntime) => Err("TryRuntime wasn't enabled when building the node. \
+				You can enable it with `--features try-runtime`."
+            .into()),
```
ETeissonniere, line 479: `Subcommand::TryRuntime(_)` could do the trick

ETeissonniere (commented):
Great work!

ETeissonniere:
@aliXsed you could add it to the GitHub actions when done, that way we always test for the migrations

node/src/cli.rs
aliXsed: This allows the command try-runtime to exist even when the feature is not enabled. In that case the command will just let the user know that they have to compile the node with the try-runtime feature enabled.

ETeissonniere: Have you tried https://github.com/NodleCode/chain/pull/444#discussion_r767384666?

aliXsed: I tested it now. This is what happens when the node is not built with the "try-runtime" and you try to call the "try-runtime" command:

> cargo run --release try-runtime                                               
    Finished release [optimized] target(s) in 0.59s
     Running `target/release/nodle-chain try-runtime`
Error: Input("TryRuntime wasn't enabled when building the node. You can enable it with `--features try-runtime`.")

Which seems what we should expect.

node/src/command.rs
aliXsed: I think the aim of the parity guys had been to have as less of the try-runtime specific types and code to be exposed for the case the feature is not enabled. If I go with `Subcommand::TryRuntime(_)`, then I have to make the type try_runtime_cli::TryRuntimeCmd which is the type what is passed to `Subcommand::TryRuntime` also available for non try-runtime builds.

aliXsed:
> @aliXsed you could add it to the GitHub actions when done, that way we always test for the migrations

@ETeissonniere This [commit](https://github.com/NodleCode/chain/pull/444/commits/7d89614f1b094f844f615933e7ee1dd218178570) should do the job for our main net. For our parachain we need to wait ro have a stable all time running collator to try the runtime upgrade on top of its state.

node/src/command.rs
```diff
@@ -436,6 +446,37 @@ pub fn run() -> Result<()> {
 
             Ok(())
         }
+        #[cfg(feature = "try-runtime")]
```
noisefloors, line 449: If we only want to build this branch on environments that are not under `test`: `#[cfg(all(not(test), feature = "try-runtime"))]. This might help with reducing overall build time per pull request.

aliXsed: As discussed on Slack, it would be a bit tricky to generalize it.

noisefloors:
The workflow related to testing is still required in order to merge according to the GitHub UI; that resolves my concerns from earlier.

aliXsed:
> The workflow related to testing is still required in order to merge according to the GitHub UI; that resolves my concerns from earlier.

True and it will be done on every pull_request and before merging anything to master.

noisefloors (approved):

node/src/cli.rs
```diff
@@ -87,6 +87,14 @@ pub enum Subcommand {
     #[structopt(name = "benchmark", about = "Benchmark runtime pallets.")]
     Benchmark(frame_benchmarking_cli::BenchmarkCmd),
 
+    /// Try some command against runtime state.
```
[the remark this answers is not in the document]
aliXsed: Done https://github.com/NodleCode/chain/pull/444/commits/34181a8a2c0ed4211f03e40a357db39ac1383f64

noisefloors (approved):
A minor change dismissed my earlier approval. It still looks good to me!

ETeissonniere (approved):

PR #445, opened 2021-12-13
refactor: improve benchmark compile output

noisefloors:
Currently, several of our benchmarks output warnings concerning unused imports in addition to the benchmark data we requested. This is unaffected by the `--quiet` flag, and in at least one case, we may be compiling more dependency code than necessary.

This PR resolves the above issues through use of Rust compiler flags.

### DIFF
--- a/pallets/allocations/src/benchmarking.rs
+++ b/pallets/allocations/src/benchmarking.rs
@@ -16,17 +16,18 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-//! Amendments pallet benchmarks
-
 #![cfg(feature = "runtime-benchmarks")]
+#![allow(unused)]
+
+//! Amendments pallet benchmarks
 
 use super::*;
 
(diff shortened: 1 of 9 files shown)


noisefloors:
This branch should be executed with the following command and return no errors in a clean environment.

```
cargo run --release --features runtime-benchmarks --manifest-path=node/Cargo.toml -- benchmark --pallet="frame_system" --extrinsic="*"
```

Note the inclusion of `--release`; this ensures we're not compiling debug or test code (which should make the runs quicker), and enable production optimizations (to place our results closer to real-world conditions).

ETeissonniere (dismissed):

noisefloors:
I resolved the errors when running `cargo test --all --all-features` by allowing unused imports in the benchmarking module specifically.

I'll merge after the test run passes.

noisefloors:
@nodle-benchbot pallet=* extrinsics=*

ETeissonniere (approved):

PR #451, opened 2022-01-06
fix(ci): remove dead code

aliXsed:
Installing ubuntu native dependencies had become dead as of a previous process simplification. It turns out CI runs fine without it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,10 +51,6 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-cargo-build-target-
 
-    - name: Install ubuntu native dependencies
-      if: matrix.os == 'ubuntu-latest'
-      run: sudo apt install -y cmake pkg-config libssl-dev git build-essential clang libclang-dev
-
     - name: Run cargo-tarpaulin
       uses: actions-rs/tarpaulin@v0.1
       with:


noisefloors (approved):

noisefloors:
It appears that the `test (ubuntu-latest)` workflow [check did pass successfully](https://github.com/NodleCode/chain/runs/4723177537?check_suite_focus=true), although GitHub isn't correctly reporting the status here. Feel free to merge. 🚀

noisefloors:
@aliXsed If you have a chance, one other change that could easily be slipped into this PR is setting our Docker upstream image to be `ubuntu:focal` instead of `ubuntu`. I suspect our current `Dockerfile` [resolves the most current release](https://wiki.ubuntu.com/Releases) (`impish`), which only has support through July, 2022 rather than April, 2025.

```
diff --git a/Dockerfile b/Dockerfile
index 7460bcfb28..7dcb1dc7ab 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu as builder
+FROM ubuntu:20.04 as builder
 
 ARG PROFILE=release
 WORKDIR /nodle-chain
@@ -15,7 +15,7 @@ RUN curl https://sh.rustup.rs -sSf | sh -s -- -y && \
 
 # ===== SECOND STAGE ======
 
-FROM ubuntu
+FROM ubuntu:20.04
```

ETeissonniere (approved):

PR #454, opened 2022-01-12
minor refactorings

ETeissonniere:
- prevent vesting to self
- remove redundant check
- fix regression test
- remove redundant variable
- cargo audit fix
- bump versions

### DIFF
--- a/pallets/grants/src/lib.rs
+++ b/pallets/grants/src/lib.rs
@@ -389,9 +389,7 @@ impl<T: Config> Pallet<T> {
         to: &T::AccountId,
         schedule: VestingScheduleOf<T>,
     ) -> DispatchResult {
-        if from == to {
-            return Err(Error::<T>::VestingToSelf.into());
-        }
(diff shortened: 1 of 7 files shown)


pallets/grants/src/lib.rs
```diff
@@ -391,6 +389,10 @@ impl<T: Config> Pallet<T> {
         to: &T::AccountId,
         schedule: VestingScheduleOf<T>,
     ) -> DispatchResult {
+        if from == to {
+            return Err(Error::<T>::VestingToSelf.into());
+        }
```
aliXsed, line 394: Can we use the `ensure` idiom here?

aliXsed:
LGTM! just left one trivial comment. Awesome.

pallets/grants/src/lib.rs
ETeissonniere: Most definitely, will fix

aliXsed (approved):

PR #510, opened 2022-02-02
Disable dependabot

ETeissonniere:
Considering how noisy it is, and that we manage the parity deps manually, disable dependabot.


noisefloors:
This seems like a good idea. As a potential alternative, we can set the [update interval to something like `monthly`](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#scheduleinterval)  to make PRs appear less often.

noisefloors (approved):

PR #524, opened 2022-02-21
feat(wnodl-pallet): be defensive against zero amount

aliXsed:
As suggested [here](https://www.notion.so/nodle/Missing-Zero-Value-Check-ce0db63556c74ff29690526f019fa979), be defensive against zero amount even when it's safe but the operation is considered useless or wasteful.

This PR contains an incidental change for min and max wrap for the development chain spec that would make the tests easier.


noisefloors:
This looks good to me. I'll follow-up once CI passes, but I don't expect it to fail.

noisefloors (approved):

PR #568, opened 2022-03-16
feat(wnodl-pallet): make origin for privillaged functions configurable

aliXsed:
Origin for initiating and settling wrapping from reserve fund will be configurable for re-users of the pallet but will keep as root for nodle-main.


runtimes/main/src/pallets_wnodl.rs
```diff
@@ -71,6 +71,7 @@ impl pallet_wnodl::Config for Runtime {
     type Currency = Balances;
     type Oracles = WnodlOracleMembership;
     type KnownCustomers = KnownCustomerMembership;
+    type ReserveTreasurerOrigin = frame_system::EnsureRoot<AccountId>;
```
noisefloors, line 74: 🔥

noisefloors (approved):

PR #569, opened 2022-03-17
productionize eden runtime

ETeissonniere:
Prepare a version of `eden` without any base call filter and ready for post migration operation:
- remove call filter
- add tech comm and mandate pallet in order to prepare for `sudo` removal

Also change a few miscellanous things:
- make sure our constants reflect a 11 decimal token instead of 12
- update our readme file to match the latest changes
- add polkadot-launch config for development purposes
- modify the development container accordingly (pre-install polkadot-launch, subwasm, and polkadot)


aliXsed (approved):
It looks very good to me. Nice Job!

PR #570, opened 2022-03-18
send allocations as a balance transfer

ETeissonniere:
Modify the way allocations are allocated to mint the coins to a special account before transfering them to the proper account. This will allow each and every allocation to show up in a `balances.Transfer` event from this same well known address (`4jbtsgNhpGAzdEGrKRb7g8Mq4ToNUpBVxeye942tWfG3gcYi`), thus making it possible for apps like Nodle Cash to fetch such events and correctly identify them as allocations on an API like Subscan.

Also modifies the way we verify the existential deposit requirements of each allocation to save on DB reads.

### DIFF
--- a/pallets/allocations/src/lib.rs
+++ b/pallets/allocations/src/lib.rs
@@ -101,6 +101,10 @@ pub mod pallet {
                 !pallet_emergency_shutdown::Pallet::<T>::shutdown(),
                 Error::<T>::UnderShutdown
             );
+            ensure!(
+                amount >= T::ExistentialDeposit::get().saturating_mul(2u32.into()),
+                Error::<T>::DoesNotSatisfyExistentialDeposit,
+            );
 
             if amount == Zero::zero() {
                 return Ok(Pays::No.into());
@@ -121,12 +125,6 @@ pub mod pallet {
             // always produce a lower number. (We use Perbill to represent percentages)
             let amount_for_protocol = T::ProtocolFee::get() * amount;
             let amount_for_grantee = amount.saturating_sub(amount_for_protocol);
-
-            Self::ensure_satisfy_existential_deposit(
-                &T::ProtocolFeeReceiver::account_id(),
-                amount_for_protocol,
-            )?;
-            Self::ensure_satisfy_existential_deposit(&to, amount_for_grantee)?;
 
             <CoinsConsumed<T>>::put(coins_that_will_be_consumed);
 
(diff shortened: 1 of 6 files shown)


pallets/allocations/src/lib.rs
```diff
@@ -83,6 +86,10 @@ pub mod pallet {
         #[pallet::weight(
 			<T as pallet::Config>::WeightInfo::allocate(proof.len() as u32)
 		)]
+        // we add the `transactional` modifier here in the event that one of the
+        // transfers fail. the code itself should already prevent this but we add
```
noisefloors, line 90: If one transfer fails, this is a preventative measure to ensure the entire batch is rolled back. Correct?

ETeissonniere: the batch would be rolled back if we use `batch_all` in the hot wallet (I believe we do?)

ETeissonniere: To clarify, I was referring to the two `transfer` calls in the function in my comment. I was initially concerned one of them would fail silently and prevent the other one from executing correctly. Which would leave us with an inconsistent state unless I applied the `#[transactional]` modifier

noisefloors: Indeed, the Hot Wallet handles transfers using `batch_all`. I verified this morning to be certain.

noisefloors (approved):

pallets/allocations/src/lib.rs
```diff
@@ -94,6 +101,10 @@ pub mod pallet {
                 !pallet_emergency_shutdown::Pallet::<T>::shutdown(),
                 Error::<T>::UnderShutdown
             );
+            ensure!(
+                amount >= T::ExistentialDeposit::get().saturating_mul(2u32.into()),
```
aliXsed, line 105: Could be nice to have the factor 2 as a defined constant here.

pallets/allocations/src/lib.rs
```diff
@@ -115,19 +126,27 @@ pub mod pallet {
...
+                // we use `KeepAlive` here because we want the guarantee that the funds left
+                // won't be considered dust, which would prevent us from sending the rest to
+                // the grantee.
+                ExistenceRequirement::KeepAlive,
+            )?;
+            T::Currency::transfer(
+                &T::PalletId::get().into_account(),
+                &to,
+                amount_for_grantee,
+                ExistenceRequirement::AllowDeath,
```
aliXsed, line 148: Perhaps we can always keep the pallet's account alive as I don't see a benefit in allowing it to die.

aliXsed:
Nice job @ETeissonniere. I left a few minor suggestions. Also if we have time we can also add another test that ensure the transfer events are being created during an allocation. This would then maintain our requirement of having those events emitted for later changes.

PR #580, opened 2022-04-07
fix: reduce maximum weight of eden to .5 seconds

ETeissonniere:
The current version of the parachain assumes a maximum block weight of 2 seconds, however, the default for parachains is actually of .5 seconds (as used by Acala, Statemine, and others). This could explain the unstable block time on eden or paradis.


rajesh-nodle:
LGTM

aliXsed (approved):

PR #586, opened 2022-04-17
technical review: fix weight computation

ETeissonniere:
SRLabs performed a technical review of our pallets and runtime. The weight functions in the `apply_as` and `apply` extrinsics is suffering from potential integer overflows and exponential complexity issues which could lead to badly computed weights/fees. This PR addresses this.


aliXsed (approved):
LGTM!

PR #592, opened 2022-04-28
Update rococo chain spec as it was restarted again. 

aliXsed:
Use the link that points to the master branch as this would be always the most updated file.


noisefloors (approved):
If it builds, ship it! :shipit:

PR #602, opened 2022-05-18
pallet-preimage integration

rajesh-nodle:
Patch to fix benchmarking weight calibration issue on all pallets with latest polkadot-v0.9.20 upgrade

### DIFF
--- a/runtimes/eden/src/lib.rs
+++ b/runtimes/eden/src/lib.rs
@@ -83,13 +83,14 @@ construct_runtime! {
 
         // Governance
         Scheduler: pallet_scheduler::{Pallet, Call, Storage, Event<T>} = 10,
-        CompanyReserve: pallet_reserve::<Instance1> = 11,
-        InternationalReserve: pallet_reserve::<Instance2> = 12,
-        UsaReserve: pallet_reserve::<Instance3> = 13,
-        Vesting: pallet_grants = 14,
-        Mandate: pallet_mandate::{Pallet, Call, Storage, Event<T>} = 15,
-        TechnicalCommittee: pallet_collective::<Instance1> = 16,
-        TechnicalMembership: pallet_membership::<Instance3> = 17,
+        Preimage: pallet_preimage::{Pallet, Call, Storage, Event<T>} = 11,
+        CompanyReserve: pallet_reserve::<Instance1> = 12,
+        InternationalReserve: pallet_reserve::<Instance2> = 13,
+        UsaReserve: pallet_reserve::<Instance3> = 14,
+        Vesting: pallet_grants = 15,
+        Mandate: pallet_mandate::{Pallet, Call, Storage, Event<T>} = 16,
+        TechnicalCommittee: pallet_collective::<Instance1> = 17,
+        TechnicalMembership: pallet_membership::<Instance3> = 18,
 
         // Consensus
         Authorship: pallet_authorship = 20,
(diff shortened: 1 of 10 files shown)


runtimes/eden/src/lib.rs
```diff
@@ -83,13 +83,14 @@ construct_runtime! {
...
-        TechnicalCommittee: pallet_collective::<Instance1> = 16,
-        TechnicalMembership: pallet_membership::<Instance3> = 17,
+        Preimage: pallet_preimage::{Pallet, Call, Storage, Event<T>} = 11,
+        CompanyReserve: pallet_reserve::<Instance1> = 12,
+        InternationalReserve: pallet_reserve::<Instance2> = 13,
+        UsaReserve: pallet_reserve::<Instance3> = 14,
+        Vesting: pallet_grants = 15,
+        Mandate: pallet_mandate::{Pallet, Call, Storage, Event<T>} = 16,
+        TechnicalCommittee: pallet_collective::<Instance1> = 17,
+        TechnicalMembership: pallet_membership::<Instance3> = 18,
```
ETeissonniere, line 93: You shouldn't change the pallet indexes as it would break compatibility with hardware wallets.

ETeissonniere (changes requested):
I am a bit unclear as to why we need this specific pallet here? I think it is a dependency for the democracy pallet but on our side it seems like a hack rather than a fix.

rajesh-nodle:
> 

benchmarking failure on `polkadot-v0.9.20` caused in the function `fn fill_schedule()` & it's looking for the interface `T::PreimageProvider`.

We decided not to use `pallet-preimage`, but looks like we need for `pallet_scheduler` context.

```
impl pallet_scheduler::Config for Runtime {
    type Event = Event;
    type Origin = Origin;
    type PalletsOrigin = OriginCaller;
    type Call = Call;
    type MaximumWeight = MaximumSchedulerWeight;
    type ScheduleOrigin = frame_system::EnsureRoot<AccountId>;
    type MaxScheduledPerBlock = MaxScheduledPerBlock;
    type OriginPrivilegeCmp = EqualPrivilegeOnly;
    type WeightInfo = pallet_scheduler::weights::SubstrateWeight<Runtime>;

    // we do not use pallets that rely on preimages at this stage
    type PreimageProvider = ();
    type NoPreimagePostponement = ();
}
```

Integrated `pallet-preimage` to make benchmarking on all pallets to work.

```
cargo run -j 14 \
	--profile release \
	--features=runtime-benchmarks \
	--manifest-path=nodes/parachain/Cargo.toml \
	-- benchmark pallet \
	--chain=eden-dev \
	--steps=50 \
	--repeat=20 \
	'--pallet=*' \
	'--extrinsic=*' \
	--execution=wasm \
	--wasm-execution=compiled \
	--heap-pages=4096 \
	--template=./.maintain/frame-weight-template.hbs \
	--output=/tmp/tmp.eGyjb4Lx2H/weights/
```

runtimes/eden/src/lib.rs
rajesh-nodle: Sure will refactor it

rajesh-nodle: Done

runtimes/eden/src/lib.rs
```diff
@@ -82,6 +82,7 @@ construct_runtime! {
         RandomnessCollectiveFlip: pallet_randomness_collective_flip = 4,
 
         // Governance
+        Preimage: pallet_preimage::{Pallet, Call, Storage, Event<T>} = 9,
```
ETeissonniere, line 85: I know I am a pain but `Preimage` probably does not fit in the `Governance` category (which also starts at `10`, not `9`). I'd suggest moving it to `System` or `Neat Things` depending on which category best fits it

ETeissonniere (approved):

noisefloors:
This branch fixed the pallet benchmark issues I was experiencing before. It built locally, but I will confirm it builds when merged to `master` before we proceed. Thank you, @rajesh-nodle!

noisefloors (approved):

PR #610, opened 2022-05-25
Remove un-needed "1 *"

aliXsed:
The recent HAL-03 suggests it.


ETeissonniere:
@aliXsed I had actually configured clippy to accept those since I found it to be more clear.

If you chose to go forward with this PR can you reconfigure clippy to warn on this kind of cases (CI GitHub action and GitHub devcontainer)

aliXsed:
@ETeissonniere Actually I'm going to close this. I did a test on https://rust.godbolt.org/ and confirmed that the compiler would optimise these things away. So no calculations actually exist on "1 *"

PR #615, opened 2022-06-12
Remove without_storage_info for the allocations pallet

rajesh-nodle:
@Team Ignore this PR, Tracking History is too High, Recreating new one

Part of [#173](https://github.com/NodleCode/chain-workspace/issues/173)

**HighLights**

- Storage `Oracles` is moved to bounded vector.
- Upperlimit  for  `Oracles`  is Configured via `MaxOracles`.
- Introduced new events `OracleMembersUpdated(u32)` & `OracleMembersOverFlow(u32, u32)`
- UnitTest Extension :: Done
- Runtime Integration :: Done
- Storage Migration (try_runtime) :: Done
- Benchmarking

**TODO**

**Ready for Intake Review**


pallets/allocations/src/tests.rs
```diff
@@ -141,7 +154,14 @@ type Events = pallet_allocations::Event<Test>;
...
 
-	let storage = frame_system::GenesisConfig::default().build_storage::<Test>().unwrap();
+	let mut storage = frame_system::GenesisConfig::default().build_storage::<Test>().unwrap();
+
+	pallet_membership::GenesisConfig::<Test> {
+		members: vec![Oracle::get()],
+		..Default::default()
+	}
+	.assimilate_storage(&mut storage)
+	.unwrap();
```
simonsso, line 164: Even though this is in the test code `unwrap()` should still be avoided, it could remove some useful clues to why a test case is failing.

simonsso (commented):
No need to change if it is complex, I just have a small unwrap allergy

pallets/allocations/src/lib.rs
```diff
@@ -67,17 +75,33 @@ pub mod pallet {
 		#[pallet::constant]
 		type ExistentialDeposit: Get<BalanceOf<Self>>;
 
+		type OracleMembersFilter: Contains<Self::AccountId>;
```
aliXsed, line 78: Trivial suggestion, feel free to skip: rename OracleMembersFilter to OracleMembers or Oracles.

pallets/allocations/Cargo.toml
```diff
@@ -16,15 +16,18 @@ std = [
   "sp-io/std",
   "sp-runtime/std",
   "sp-std/std",
+  "pallet-membership/std",
```
aliXsed, line 19: We shouldn't need a dependency to `pallet-membership`, for the tests and as dev dependency it would be totally okay.

pallets/allocations/src/lib.rs
```diff
@@ -169,18 +193,14 @@ pub mod pallet {
 		DoesNotSatisfyExistentialDeposit,
 	}
 
-	#[pallet::storage]
-	#[pallet::getter(fn oracles)]
-	pub type Oracles<T: Config> = StorageValue<_, Vec<T::AccountId>, ValueQuery>;
-
```
aliXsed, line 175: Glad to see it go. Good job @rajesh-nodle

pallets/allocations/src/migrations.rs
```diff
@@ -0,0 +1,162 @@
...
+				current,
+				onchain
+			);
+
+			if current == 1 && onchain == 0 {
+				let pallet_prefix: &[u8] = b"Allocations";
+				let storage_item_prefix: &[u8] = b"Oracles";
+
+				if have_storage_value(pallet_prefix, storage_item_prefix, &[]) {
+					remove_storage_prefix(pallet_prefix, storage_item_prefix, &[]);
```
aliXsed, line 47: 👍  for removing the redundant Oracles storage.

pallets/allocations/src/migrations.rs
```diff
@@ -0,0 +1,162 @@
...
+
+					current.put::<Pallet<T>>();
```
aliXsed, line 49: Can you please take an explicit approach similar to https://github.com/paritytech/substrate/blob/polkadot-v0.9.20/frame/staking/src/migrations.rs#L50 to write the new storage version on chain here?

pallets/allocations/src/tests.rs
```diff
@@ -156,12 +156,19 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
...
+	let rval = pallet_membership::GenesisConfig::<Test> {
 		members: vec![Oracle::get()],
 		..Default::default()
 	}
-	.assimilate_storage(&mut storage)
-	.unwrap();
+	.assimilate_storage(&mut storage);
+
+	if rval.is_err() {
+		log::error!(
```
simonsso: `panic` as you have in the other version is better than `log::error!()` which will only print an error message without chaining the return codes

pallets/allocations/src/migrations.rs
```diff
@@ -0,0 +1,162 @@
...
+				line!(),
+				current,
+				onchain
+			);
+
+			if current == 1 && onchain == 1 {
+				let pallet_prefix: &[u8] = b"Allocations";
+				let storage_item_prefix: &[u8] = b"Oracles";
+
+				if get_storage_value::<Vec<T::AccountId>>(pallet_prefix, storage_item_prefix, &[]).is_none() {
```
aliXsed, line 132: Post migration you cannot do this test as `Oracles` is removed, isn't it? In that case please feel free to remove this check and no need to store the pre-migration numbers in pre-upgrade either

pallets/allocations/Cargo.toml
```diff
@@ -35,8 +38,10 @@ pallet-balances = { git = "https://github.com/paritytech/substrate", default-fea
 sp-io = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.20" }
 sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.20" }
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.20" }
+pallet-membership = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.20" }
```
aliXsed, line 41: I suggest moving this to dev-dependencies

pallets/allocations/src/lib.rs
rajesh-nodle: Done

pallets/allocations/Cargo.toml
rajesh-nodle: Done

pallets/allocations/src/migrations.rs
rajesh-nodle: Migrated handling of `StorageVersion`, similar to shared substrate `staking` pallet.

pallets/allocations/src/tests.rs
rajesh-nodle: Done

pallets/allocations/src/migrations.rs
rajesh-nodle: Here we are ensuring storage `Oracles` is removed. I hope better to ensure that state. As suggested removed the temp storage of Oracles count.

rajesh-nodle:
@Team, benchmarking done, but by adding support for Instance on `pallet-allocations`, but this dependency is coming due to below expression.

Where we are injecting the address to `pallet-members` before doing the benchmark.



Let me know if there is better alternate.

aliXsed:
> @Team, benchmarking done, but by adding support for Instance on `pallet-allocations`, but this dependency is coming due to below expression.
> 
> Where we are injecting the address to `pallet-members` before doing the benchmark.
> 
> 
> 
> Let me know if there is better alternate.

@rajesh-nodle In my opinion, we shouldn't need to make the allocations pallet a multiple instance pallet or add any extra assumption like a dependency to the membership pallet just for the sake of benchmarking. I highly suggest you look and take the same approach as once taken by wnodl: https://github.com/NodleCode/chain/blob/2.0.18/pallets/wnodl/src/lib.rs#L103-L107
This way we can avoid the huge change and unnecessary complexity of handling the instance `I`. When you do that the pallet configuration would be as simple as:
```
	pub trait Config: frame_system::Config
	{
		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
        ...
```

rajesh-nodle:
Replaced with https://github.com/NodleCode/chain/pull/634

PR #618, opened 2022-06-12
Remove without_storage_info for the grants pallet

rajesh-nodle:
@Team Ignore this PR, Tracking History is too High, Recreating new one

Part of [#173](https://github.com/NodleCode/chain-workspace/issues/173)

**HighLights**

- Storage `VestingSchedules` is moved to bounded vector.
- Upperlimit  for  `VestingSchedules`  is Configured via `MaxSchedule`.
- Introduce error code `MaxScheduleOverflow`.
- UnitTest Extension :: Done
- Runtime Integration :: Done
- Storage Migration (try_runtime) :: Done
- Benchmarking

**TODO**

**Ready for Intake Review**


pallets/grants/src/lib.rs
```diff
@@ -216,16 +220,19 @@ pub mod pallet {
 			log::debug!("got new schedules for overwriting: {:#?}", new_schedules);
 
 			let target = T::Lookup::lookup(who)?;
-			VestingSchedules::<T>::insert(target.clone(), new_schedules.clone());
+
+			let new_schedules_bvec =
+				<BoundedVec<VestingScheduleOf<T>, T::MaxSchedule>>::try_from(new_schedules.clone())
+					.map_err(|_| <Error<T>>::MaxScheduleOverflow)?;
+
+			VestingSchedules::<T>::insert(target.clone(), new_schedules_bvec);
```
ETeissonniere, line 228: If you set the `new_schedules` param to be a `BoundedVec` you should be able to avoid this.

pallets/grants/src/lib.rs
```diff
@@ -315,7 +323,15 @@ pub mod pallet {
 
 				T::Currency::resolve_creating(who, T::Currency::issue(total_grants));
 				T::Currency::set_lock(VESTING_LOCK_ID, who, total_grants, WithdrawReasons::all());
-				<VestingSchedules<T>>::insert(who, schedules);
+
+				match <BoundedVec<VestingScheduleOf<T>, T::MaxSchedule>>::try_from(schedules.clone()) {
```
aliXsed, line 327: Similar to what Eliot suggested, can we collect the schedules of someones grants in BoundedVec rather than Vec at line 305? If so, then this step of converting vec to bounded vec could be avoided.

pallets/grants/src/lib.rs
```diff
@@ -103,7 +105,7 @@ impl<BlockNumber: AtLeast32Bit + Copy, Balance: AtLeast32Bit + Copy> VestingSche
 #[frame_support::pallet]
 pub mod pallet {
 	use super::*;
-	use frame_support::pallet_prelude::*;
+	// use frame_support::pallet_prelude::*;
```
aliXsed, line 108: Can we remove this?

pallets/grants/src/lib.rs
```diff
@@ -377,7 +391,12 @@ impl<T: Config> Pallet<T> {
 
 		T::Currency::transfer(from, to, schedule_amount, ExistenceRequirement::AllowDeath)?;
 		T::Currency::set_lock(VESTING_LOCK_ID, to, total_amount, WithdrawReasons::all());
-		<VestingSchedules<T>>::mutate(to, |v| (*v).push(schedule));
+		<VestingSchedules<T>>::mutate(to, |vesting_schedules| -> DispatchResult {
+			vesting_schedules
+				.try_push(schedule)
+				.map_err(|_| <Error<T>>::MaxScheduleOverflow)?;
+			Ok(())
+		})?;
```
aliXsed, line 399: The error should be checked and covered before `transfer` and `set_lock`. When there is an error we should make sure we haven't changed the storage.

runtimes/eden/src/pallets_util.rs
```diff
@@ -29,11 +29,16 @@ use frame_system::{EnsureRoot, EnsureSigned};
 use primitives::{AccountId, Balance};
 use sp_runtime::Perbill;
 
+parameter_types! {
+	pub const MaxSchedule: u32 = 200;
```
aliXsed, line 33: Before finalising this number, can we know what's the highest number of the schedules currently in our Eden storage?

pallets/grants/src/lib.rs
rajesh-nodle: Indeed not relevant function `overwrite_vesting_schedules()` got removed.

pallets/grants/src/lib.rs
rajesh-nodle: Sure

pallets/grants/src/lib.rs
rajesh-nodle: Refactored the code, execution order got changed.

But, extrinsics behaviour is, All the modified storage states 
are restored back if there is any error in transactions.

runtimes/eden/src/pallets_util.rs
rajesh-nodle: Yes right now this is stub value, will get the actual as suggested

pallets/grants/src/lib.rs
rajesh-nodle: Can I have bit more details here ?

runtimes/eden/src/pallets_util.rs
rajesh-nodle: From the existing state of `paradis`, Max length of Vesting schedule is 10, right now configured to double the size 20.

pallets/grants/src/lib.rs
```diff
@@ -288,34 +267,26 @@ pub mod pallet {
...
+				let vesting_schedule: BoundedVec<VestingScheduleOf<T>, T::MaxSchedule> = schedules
+					.iter()
+					.map(|&(start, period, period_count, per_period)| VestingSchedule {
+						start,
+						period,
+						period_count,
+						per_period,
+					})
+					.collect::<Vec<_>>()
+					.try_into()
```
aliXsed, line 280: How about going for a bounded vec at line 255 where `GenesisConfig` is defined to avoid converting from Vec to BoundedVec here?

pallets/grants/src/lib.rs
```diff
@@ -375,9 +344,16 @@ impl<T: Config> Pallet<T> {
 			.checked_add(&schedule_amount)
 			.ok_or(Error::<T>::NumOverflow)?;
 
-		T::Currency::transfer(from, to, schedule_amount, ExistenceRequirement::AllowDeath)?;
-		T::Currency::set_lock(VESTING_LOCK_ID, to, total_amount, WithdrawReasons::all());
-		<VestingSchedules<T>>::mutate(to, |v| (*v).push(schedule));
+		<VestingSchedules<T>>::try_mutate(to, |vesting_schedules| -> DispatchResult {
+			vesting_schedules
+				.try_push(schedule)
+				.map_err(|_| <Error<T>>::MaxScheduleOverflow)?;
```
aliXsed, line 350: How about moving this to line 343 before calculating total_amount so we don't calculate total_amount if we there is vesting schedule vector is going to overflow.

pallets/grants/src/migrations.rs
```diff
@@ -0,0 +1,199 @@
...
+				let pallet_prefix: &[u8] = b"Vesting";
+				let storage_item_prefix: &[u8] = b"VestingSchedules";
+
+				// Check number of entries, and set it aside in temp storage
+				let stored_data: Vec<_> =
+					storage_key_iter::<T::AccountId, Vec<VestingScheduleOf<T>>, Blake2_128Concat>(
+						pallet_prefix,
+						storage_item_prefix,
+					)
+					.collect();
```
aliXsed, line 128: I would suggest that after this stage we go through all the stored_date and check none of the accounts have grants more than the MaxSchedule. If we can show that in `pre-upgrade`, we can be confident `on_runtime_upgrade` is not going to fail. It makes on_runtime_upgrade safer because a failure would be flagged before making any changes.

pallets/grants/src/lib.rs
aliXsed: Resolving this, as there has been changes to this file.

runtimes/eden/src/pallets_util.rs
aliXsed: @ETeissonniere as Rajesh suggests we can be safe with 20 for the maximum number of grants for individual accounts based on what we have currently stored. However I would like to know your intuition on what would be a good cap here?

aliXsed:
@rajesh-nodle This PR looks very good to me. I have just left a few suggestions for you. I believe we can get this merged very soon.

pallets/grants/src/lib.rs
rajesh-nodle: Looks like we can't do it ...

* Due to `trait bound constratints on serde`. And all the pallets in substrate are using `Vec<>` inside `pallet::genesis_config`, and there is no `derive` appart from `#[cfg_attr(feature = "std", derive(Default))]`

* Secondly the `type` hold inside `Vec` & `BoundedVec` containers are different. type transformation is required.

pallets/grants/src/migrations.rs
rajesh-nodle: Done

pallets/grants/src/migrations.rs
```diff
@@ -130,6 +130,10 @@ pub mod v1 {
 				let mapping_count = stored_data.len();
 				Self::set_temp_storage(mapping_count as u32, "mapping_count");
 
+				stored_data
+					.iter()
+					.for_each(|(_account, old_vesting)| assert!(old_vesting.len() as u32 <= T::MaxSchedule::get()));
```
simonsso, line 135: There is a theoretical chance `len()` as a `usize` is larger than an `u32` I think the cast should be to extend the other side to `usize` is better.

rajesh-nodle: Done

rajesh-nodle:
For the comment ...

image

Yes true `GenesisBuild` gets called during launch for example `fn eden_testnet_genesis()` from the node client. But not during the Runtime Upgrade.

pallets/grants/src/lib.rs
simonsso: Is it possible to check with a  `schedule.len()` which should be a quick operation?

pallets/grants/src/migrations.rs
```diff
@@ -0,0 +1,203 @@
...
+					storage_key_iter::<T::AccountId, Vec<VestingScheduleOf<T>>, Blake2_128Concat>(
+						pallet_prefix,
+						storage_item_prefix,
+					)
+					.collect();
+
+				let mut translated = 0u64;
+				let mut max_schedules: usize = 0;
+
+				assert!(stored_data.len() > 0);
```
simonsso, line 61:
```suggestion
				assert!(!stored_data.is_empty());
```

Reading this warning in full and it makes sense

```
Warning:   --> pallets/grants/src/migrations.rs:61:13
   |
61 |                 assert!(stored_data.len() > 0);
   |                         ^^^^^^^^^^^^^^^^^^^^^ help: using `!is_empty` is clearer and more explicit: `!stored_data.is_empty()`
   |
   = note: `#[warn(clippy::len_zero)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_zero
```

rajesh-nodle: Done

pallets/grants/src/migrations.rs
```diff
@@ -0,0 +1,203 @@
...
+
+				let stored_data: Vec<_> = storage_key_iter::<
+					T::AccountId,
+					BoundedVec<VestingScheduleOf<T>, T::MaxSchedule>,
+					Blake2_128Concat,
+				>(pallet_prefix, storage_item_prefix)
+				.collect();
+
+				let mapping_count = stored_data.len() as u32;
+
```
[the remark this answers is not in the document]
rajesh-nodle: we are using it in below expression

pallets/grants/src/migrations.rs
```diff
@@ -0,0 +1,203 @@
...
+				log::info!(
+					"post_upgrade[{:#?}]=> VestingSchedules map count :: [{:#?}]",
+					line!(),
+					mapping_count,
+				);
+
+				// Check number of entries matches what was set aside in pre_upgrade
+				if let Some(pre_mapping_count) = Self::get_temp_storage::<u32>("mapping_count") {
```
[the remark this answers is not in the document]
rajesh-nodle: Good one, Done

pallets/grants/src/migrations.rs
```diff
@@ -0,0 +1,205 @@
...
+
+				// // Check number of entries matches what was set aside in pre_upgrade
+				// if let Some(pre_mapping_count) = Self::get_temp_storage::<u32>("mapping_count") {
+				// 	assert!(pre_mapping_count == mapping_count);
+				// } else {
+				// 	log::info!(
+				// 		"post_upgrade[{:#?}]=> Pre-Migration did not executed. This probably should be removed",
+				// 		line!(),
+				// 	);
+				// }
```
aliXsed, line 194: Do we need to keep the commented lines?

pallets/grants/src/migrations.rs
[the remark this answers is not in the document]
aliXsed: Looks like you don't need to `collect` the iterator if your only use case is checking the length. There seems to be a `count` method for the iterator for this purpose.

pallets/grants/src/lib.rs
aliXsed: @simonsso If we could check the `schedule.len()` and then do `push` rather than `try_push`, then your suggestion could be possibly better. However on a bounded vector you can only do `try_push` meaning we still need to cover the error case here anyway.

pallets/grants/src/migrations.rs
rajesh-nodle: no not required, statements are optimised.

pallets/grants/src/migrations.rs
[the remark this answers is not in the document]
rajesh-nodle: We need the storage content to ensure old vesting schedule length are below `T::MaxSchedule`

image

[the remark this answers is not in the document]
rajesh-nodle: Got it, comment is on `post_upgrade()`, Done.

PR #626, opened 2022-06-24
allocations optim

ETeissonniere:
Attempt to optimize the allocations pallet:
- save 1 write by checking our issuance against our hardcoded maximum supply instead of tracking coins issued over time.
- save 1 read by removing the emergency shutdown feature which wasn't used and could be replaced by the removal of the instant oracle.
- save some non-quantifiable resources and indexing work for Subscan by removing our superfluous event at the end. We now rely on the two `Transfer` events from the `transfer` calls for indexing in both Subscan and the Nodle App.
- saves hundreds of reads and writes by using a properly optimized batching extrinsic appropriately named `batch` which will allow us to be much more efficient in batching allocations together VS our current approach of using `utility.batch_all`. Deprecate `allocate` accordingly.

When reviewing, pay attention to:
- the issuance checks, we do not want to have an in-prod problem because I messed up my computation here!
- what other smart optimizations can we do without modifying the pallet's behavior? (as in, we still need our two transfer calls at the end for Subscan to pick allocations up)


pallets/allocations/src/lib.rs
```diff
@@ -93,13 +100,9 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			to: T::AccountId,
 			amount: BalanceOf<T>,
-			proof: Vec<u8>,
+			_proof: Vec<u8>,
```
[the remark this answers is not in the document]
ETeissonniere: because the team still gives/wants a proof to link every rewards to some 'work' in the backend

pallets/allocations/src/lib.rs
```diff
@@ -109,13 +112,9 @@ pub mod pallet {
...
-			let coins_that_will_be_consumed = coins_already_allocated
-				.checked_add(&amount)
-				.ok_or("Overflow computing coins consumed")?;
-
+			let current_supply = T::Currency::total_issuance();
 			ensure!(
-				coins_that_will_be_consumed <= T::MaximumCoinsEverAllocated::get(),
+				current_supply.saturating_add(amount) <= T::MaximumSupply::get(),
 				Error::<T>::TooManyCoinsToAllocate
 			);
```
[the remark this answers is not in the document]
ETeissonniere: I would love a per block, or per time window limit. Though it would be more of a new feature VS an optimization.

The original thinking (re-issuance) was to provide a guarantee we'd never go over 21B. I had originally computed the "magic value" `MaximumCoinsEverAllocated` myself to guarantee this, the current changes simply make it more expressive, and easier to verify

pallets/allocations/src/lib.rs
```diff
@@ -61,12 +62,16 @@ pub mod pallet {
...
-		type MaximumCoinsEverAllocated: Get<BalanceOf<Self>>;
+		type MaximumSupply: Get<BalanceOf<Self>>;
 
 		/// Runtime existential deposit
 		#[pallet::constant]
 		type ExistentialDeposit: Get<BalanceOf<Self>>;
 
+		/// How big a batch can be
+		#[pallet::constant]
+		type MaxAllocs: Get<u32>;
```
aliXsed, line 73: Nice that we have kept it easily configurable.

pallets/allocations/src/lib.rs
```diff
@@ -77,84 +82,99 @@ pub mod pallet {
...
-		#[pallet::weight(
-			<T as pallet::Config>::WeightInfo::allocate(proof.len() as u32)
-		)]
-		// we add the `transactional` modifier here in the event that one of the
-		// transfers fail. the code itself should already prevent this but we add
-		// this as an additional guarantee.
+		/// Optimized allocation call, which will batch allocations of various amounts
+		/// and destinations and together. This allow us to be much more efficient and thus
+		/// increase our chain's capacity in handling these transactions.
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::batch(batch.len().try_into().unwrap_or(T::MaxAllocs::get())))]
```
aliXsed, line 97: I think `batch.len() as u32` is a pretty safe cast here especially that the batch is bounded to MaxAlloc which is Get<u32>.

pallets/allocations/src/lib.rs
```diff
@@ -77,84 +82,99 @@ pub mod pallet {
...
-				Error::<T>::DoesNotSatisfyExistentialDeposit,
-			);
 
-			if amount == Zero::zero() {
-				return Ok(Pays::No.into());
+			// sanity checks
+			let mut full_issuance: BalanceOf<T> = Zero::zero();
+			for (_account, amount) in batch.iter().cloned() {
+				ensure!(
+					amount >= T::ExistentialDeposit::get().saturating_mul(2u32.into()),
```
aliXsed, line 109: Not related to this PR, but I wonder why we have checked 2 x ED here? We probably don't need to check the `amount` at all because ED would only matter for to the resulted balance and not the amount we are allocating. If we need to have a minimum allocation, it would be better to be like `MaxAllocs` configurable.

pallets/allocations/src/lib.rs
```diff
@@ -77,84 +82,99 @@ pub mod pallet {
...
-			ensure!(
-				amount >= T::ExistentialDeposit::get().saturating_mul(2u32.into()),
-				Error::<T>::DoesNotSatisfyExistentialDeposit,
-			);
 
-			if amount == Zero::zero() {
-				return Ok(Pays::No.into());
+			// sanity checks
+			let mut full_issuance: BalanceOf<T> = Zero::zero();
+			for (_account, amount) in batch.iter().cloned() {
```
aliXsed, line 107: Can we avoid "cloning" here, as what we are doing is just checking and not making any change to the batch.

aliXsed, line 107: A fold or reduce pattern seem to be a nice match for what we are doing here.

pallets/allocations/src/lib.rs
```diff
@@ -77,84 +82,99 @@ pub mod pallet {
...
+					.checked_add(&amount)
+					.ok_or(Error::<T>::TooManyCoinsToAllocate)?;
 			}
 
-			let coins_already_allocated = Self::coins_consumed();
-			let coins_that_will_be_consumed = coins_already_allocated
-				.checked_add(&amount)
-				.ok_or("Overflow computing coins consumed")?;
+			if full_issuance == Zero::zero() {
+				return Ok(Pays::No.into());
```
aliXsed, line 120: Would returning an error here help our oracle to notice a bug/unexpected behaviour on their side?

pallets/allocations/src/lib.rs
```diff
@@ -77,84 +82,99 @@ pub mod pallet {
...
+			}
 
+			let current_supply = T::Currency::total_issuance();
 			ensure!(
-				coins_that_will_be_consumed <= T::MaximumCoinsEverAllocated::get(),
+				current_supply.saturating_add(full_issuance) <= T::MaximumSupply::get(),
```
aliXsed, line 125: How about a gradually more aggressive control on the issuance through:
```suggestion
				full_issuance <= T::MaximumSupply::get().saturating_sub(current_supply).div(2u32.into()),
```
I'm just hoping that we can make this check more useful. If you take this suggestion then `use sp_std::ops::Div;` too.

pallets/allocations/src/lib.rs
```diff
@@ -77,84 +82,99 @@ pub mod pallet {
...
+			for (account, amount) in batch.iter().cloned() {
+				let amount_for_protocol = T::ProtocolFee::get() * amount;
+				let amount_for_grantee = amount.saturating_sub(amount_for_protocol);
+				T::Currency::transfer(
+					&T::PalletId::get().into_account(),
+					&account,
+					amount_for_grantee,
+					ExistenceRequirement::KeepAlive,
+				)?;
+				full_protocol = full_protocol.saturating_add(amount_for_protocol);
```
aliXsed, line 143: Nice accumulation here. I bet it has caused at least 2x efficiency.

pallets/allocations/src/lib.rs
```diff
@@ -77,84 +82,99 @@ pub mod pallet {
...
+		#[deprecated(note = "allocate is sub-optimized and chain heavy")]
+		pub fn allocate(
+			origin: OriginFor<T>,
+			to: T::AccountId,
+			amount: BalanceOf<T>,
+			_proof: Vec<u8>,
+		) -> DispatchResultWithPostInfo {
+			let one_batch: BoundedVec<(T::AccountId, BalanceOf<T>), T::MaxAllocs> = vec![(to, amount)]
+				.try_into()
+				.expect("one element should fit inside the bounded vec; qed");
```
aliXsed, line 175: How about
```suggestion
			let one_batch: BoundedVec<(T::AccountId, BalanceOf<T>), T::MaxAllocs> = bounded_vec![(to, amount)];
```
If you take this suggestion then `use frame_support::bounded_vec;` too.

pallets/allocations/src/tests.rs
```diff
@@ -38,10 +37,9 @@ frame_support::construct_runtime!(
 		NodeBlock = Block,
 		UncheckedExtrinsic = UncheckedExtrinsic,
 	{
-		System: frame_system::{Pallet, Call, Config, Storage, Event<T>},
-		Balances: pallet_balances::{Pallet, Call, Config<T>, Storage, Event<T>},
-		EmergencyShutdown: pallet_emergency_shutdown::{Pallet, Call, Storage, Event<T>},
```
aliXsed, line 43: Always happy to see eliminating any wastes. Thank you.

pallets/allocations/src/lib.rs
ETeissonniere: yup, most definitely

pallets/allocations/src/lib.rs
ETeissonniere: Why are we dividing by 2 here?

aliXsed: It can be div by any number, but 2 happens to be implemented a lot faster most of the time. The idea is at any single allocation we can only go for a portion of the available supply and not all or close to all.

pallets/allocations/src/lib.rs
ETeissonniere: sounds like we can't cast blindly as a u32 based https://stackoverflow.com/questions/29592256/whats-the-difference-between-usize-and-u32

pallets/allocations/src/lib.rs
ETeissonniere: I would rather not do this, or at least not in this PR, since this pallet is supposed to be the only one to issue coins I do not really see the advantage of limiting it in such a way. If we want a cap on allocs (which we probably should), it sounds like something that we'd rather implement in a separate PR, and probably in a very different way (per window limit for instance)

aliXsed: Fair enough. We can improve this later with a dedicated PR and I'm also on the side of setting individual window based limits for both the oracle and the destination account.

aliXsed (dismissed):
LGTM and multiple times more efficient than our current implementation, So it's a go from me.

pallets/allocations/src/benchmarking.rs
```diff
@@ -26,37 +26,51 @@ use super::*;
...
 benchmarks! {
 	allocate {
-		let b in 1 .. MAX_BYTES;
+		let config = make_benchmark_config::<T>();
+
+		Pallet::<T>::initialize_members(&[config.oracle.clone()]);
+	}: _(RawOrigin::Signed(config.oracle.clone()), config.grantee.clone(), T::ExistentialDeposit::get() * 10u32.into(), vec![])
+
+	batch {
+		let b in 1..T::MaxAllocs::get();
```
simonsso, line 67: When a range starts at 1 and does not include the last element it looks suspicious, is this intended here?

pallets/allocations/src/lib.rs
```diff
@@ -22,17 +22,19 @@ mod benchmarking;
...
 use frame_support::{
+	dispatch::Weight,
 	ensure,
+	migration::remove_storage_prefix,
 	traits::{tokens::ExistenceRequirement, ChangeMembers, Currency, Get, InitializeMembers},
 	transactional, PalletId,
 };
 use frame_system::ensure_signed;
-use sp_runtime::traits::AccountIdConversion;
+use sp_runtime::traits::{AccountIdConversion, CheckedAdd};
```
simonsso, line 32: Join all with line 37

pallets/allocations/src/benchmarking.rs
ETeissonniere: actually a batch len of 0, we'd throw an error since there are no allocations to perform

simonsso: And the last to call should be `T::MaxAllocs::get()` or `T::MaxAllocs::get()-1`

PR #629, opened 2022-07-07
Silence a few clippy warnings

simonsso:
Should make the build clean.

Removed `1 * something`  it existed only for consistency with other values in the lists

`derive_partial_eq_without_eq` implemented suggestion but another source was included in the same file so I also silenced it

`#[allow(clippy::type_complexity)]` yes the types are complex let's just acknowledge that

`#[allow(clippy::boxed_local)]` not sure if we should change the code or silence the warning so I silenced it.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -15,6 +15,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+#![allow(clippy::derive_partial_eq_without_eq)]
 
 use cumulus_primitives_core::ParaId;
 use primitives::{AccountId, Balance, Signature};
@@ -39,7 +40,7 @@ pub fn get_public_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pa
 }
 
 /// The extensions for the [`ChainSpec`].
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ChainSpecGroup, ChainSpecExtension)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, ChainSpecGroup, ChainSpecExtension)]
 #[serde(rename_all = "camelCase")]
 pub struct Extensions {
 	/// The relay chain of the Parachain.
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -61,6 +61,7 @@ type HostFunctions = (
 ///
 /// Use this macro if you don't actually need the full service, but just the builder in order to
 /// be able to perform chain operations.
+#[allow(clippy::type_complexity)]
 pub fn new_partial<RuntimeApi, BIQ>(
 	config: &Configuration,
 	build_import_queue: BIQ,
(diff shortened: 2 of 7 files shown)


runtimes/eden/src/pallets_system.rs
```diff
@@ -102,7 +102,7 @@ impl pallet_balances::Config for Runtime {
 }
 
 parameter_types! {
-	pub const TransactionByteFee: Balance = 1 * constants::MICRO_NODL;
+	pub const TransactionByteFee: Balance = constants::MICRO_NODL;
```
aliXsed, line 105: I had suggested this change before. The thing is in this case clippy seems to be wrong. There is no extra overhead or unnecessary multiplication after this line is compiled, I have tested it. Then it goes back to the matter of style, which some prefer consistency (so this line looks similar to the previous lines), some would prefer conciseness. I'm not strongly in favour or against any of these two. I like to not see the clippy warning any way.

runtimes/eden/src/pallets_util.rs
```diff
@@ -105,7 +105,7 @@ parameter_types! {
 	pub const ClassDeposit: Balance = 10 * constants::NODL;
 	pub const InstanceDeposit: Balance = 100 * constants::MILLI_NODL;
 	pub const MetadataDepositBase: Balance = 10 * constants::NODL;
-	pub const MetadataDepositPerByte: Balance = 1 * constants::NODL;
+	pub const MetadataDepositPerByte: Balance = constants::NODL;
```
aliXsed, line 108: I had suggested this change before. The thing is in this case clippy seems to be wrong. There is no extra overhead or unnecessary multiplication after this line is compiled, I have tested it. Then it goes back to the matter of style, which some prefer consistency (so this line looks similar to the previous lines), some would prefer conciseness. I'm not strongly in favour or against any of these two. I like to not see the clippy warning any way.

node/src/service.rs
```diff
@@ -61,6 +61,7 @@ type HostFunctions = (
 ///
 /// Use this macro if you don't actually need the full service, but just the builder in order to
 /// be able to perform chain operations.
+#[allow(clippy::type_complexity)]
```
aliXsed, line 64: This line falls into one of the files that we need to keep updated with the upstream. We have a constant burden upgrading our code to be compatible with the upstream. So in general I'm in favour of making minimal changes to such places to make the upgrade easier.

aliXsed:
@simonsso Thank you for making the effort to remove the clippy warnings.

runtimes/eden/src/pallets_util.rs
simonsso: I can silence this warning insted

node/src/chain_spec.rs
```diff
@@ -15,6 +15,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+#![allow(clippy::derive_partial_eq_without_eq)]
```
aliXsed, line 18: Now that `Eq` is derived for `Extensions`, do we still need to silence clippy for this warning?

node/src/service.rs
simonsso: @aliXsed we could try to get in included upstream, where does this file come from? I did not find it on master in substrate

node/src/chain_spec.rs
simonsso: With our current master it is still needed.

simonsso:
I think this is ready to be merged, please review.

aliXsed (approved):

PR #637, opened 2022-07-13
Fix to CI/CD build try-runtime error

rajesh-nodle:
`try-runtime` target needs +nightly toolchain to be default ...


noisefloors (approved):
Setting the installation as default seems to have worked. LGTM.

PR #643, opened 2022-07-21
fix(pallet-allocations): fix try-runtime and the migration code

aliXsed:
Remove the attempt to remove CoinsConsumed which was already removed in a previous deployment of release 2.0.21. Fix the try-runtime test which could have indicated this error to us before merging the migration code.

### DIFF
--- a/pallets/allocations/src/lib.rs
+++ b/pallets/allocations/src/lib.rs
@@ -55,13 +55,13 @@ type BalanceOf<T> = <<T as Config>::Currency as Currency<<T as frame_system::Con
 // migration logic. This should match directly with the semantic versions of the Rust crate.
 #[derive(Encode, Decode, MaxEncodedLen, Clone, Copy, PartialEq, Eq, RuntimeDebug, TypeInfo)]
 enum Releases {
-	V0_0_0Legacy, // To handle Legacy version
-	V2_0_21,
+	V0, // Legacy version
+	V1, // Adds storage info
 }
 
 impl Default for Releases {
 	fn default() -> Self {
-		Releases::V0_0_0Legacy
+		Releases::V0
 	}
 }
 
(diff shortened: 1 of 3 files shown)


simonsso (approved):

pallets/allocations/src/lib.rs
```diff
@@ -55,21 +55,20 @@ type BalanceOf<T> = <<T as Config>::Currency as Currency<<T as frame_system::Con
 // migration logic. This should match directly with the semantic versions of the Rust crate.
 #[derive(Encode, Decode, MaxEncodedLen, Clone, Copy, PartialEq, Eq, RuntimeDebug, TypeInfo)]
 enum Releases {
-	V0_0_0Legacy, // To handle Legacy version
-	V2_0_21,
+	V0, // Legacy version
+	V1, // Adds storage info
```
rajesh-nodle, line 59: Is it not right way to enumerate by Pallet version ?

pallets/allocations/src/migrations.rs
```diff
@@ -18,128 +18,92 @@
 
 pub mod v1 {
 	use crate::{Config, Releases, StorageVersion};
+	use frame_support::{storage::migration::remove_storage_prefix, traits::Get, weights::Weight};
 
-	#[cfg(feature = "try-runtime")]
-	use crate::BalanceOf;
+	pub fn on_runtime_upgrade<T: Config>() -> Weight {
```
rajesh-nodle, line 23: Can we take this as suggested migration immplementation pattern and apply to other PR's ?

rajesh-nodle (commented):
LGTM, Added some clarification questions on implementation.

pallets/allocations/src/migrations.rs
aliXsed: I would suggest it

PR #660, opened 2022-08-17
Fredrik/polkadot_v0.9.27

simonsso:
Upgrade to polkadot v0.9.27

Build ok
Benchmarks run
Try runitme green
Local testnet starting.

Ready for testnet.


pallets/allocations/src/tests.rs
```diff
@@ -267,4 +267,4 @@ fn no_issuance() {
 			Errors::BatchEmpty
 		);
 	})
-}
```
simonsso, line 270: I think this was added by `cargo fmt`, and I agree, I think we should have newlines at `EOF`

aliXsed (approved):
LGTM

ETeissonniere:
Amazing! Let's make sure Zondax and Mobile run their tests once it lands on the testnet as usual

PR #662, opened 2022-08-22
renounce privileges

ETeissonniere:
Modify the `grants` pallet so that the `CancelOrigin` can renounce its
privileges to call `cancel_all_vesting_schedules` for a given account.


aliXsed:
This feature could be a lot more useful if the origin of the root level operations which goes through pallet-mandate was configured to be a different collection of wallets or at least a higher threshold of yes (e.g. 3/4 vs 1/2) was required for mandating an upgrade. However with the current number of technical committee members, A 3/4th threshold would allow a mere two wallets to obstruct any future amendments which doesn't seem to be safe. So I'm leaning towards yes to this feature as it is and accept its limitation. With an improved government structure later, we can make sure that different privileged operations have different origins, when it makes them more reasonable or safer.

pallets/grants/src/lib.rs
```diff
@@ -261,6 +259,10 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn renounced)]
+	pub type Renounced<T: Config> = CountedStorageMap<_, Blake2_128Concat, T::AccountId, bool, ValueQuery>;
```
aliXsed, line 264: Just wondering why we'd need a `CountedStorageMap` here? Could a`StorageMap` work as well, considering it's a bit lighter too.

ETeissonniere: Honestly copy and pasted this from the vesting schedule storage map. Hadn't realized it wasn't a pure StorageMap. Will fix

aliXsed (approved):
LGTM

PR #667, opened 2022-09-08
feat: limit allocations to targeted supply inflation

aliXsed:
The mint curve limiter is just an upper bound enforced for security and commitment to Nodle's tokenomics. The actual mint depends on the allocation oracles and the community's growth and participation. 

**Note 1**: to find out what does `Perbill::from_perthousand(1)` as an inflation rate step for Eden mean, you need to do the following calculation: 
Suppose the total issuance at the beginning of that quarter is around 8.4B NODLs. Then the allowed increase on the supply would be as follows `8.4B * 0.001 = 8.4M`.  Since the fiscal period for Eden's configuration is based on 90 days. Then 8.4M increase is for that period. The actual limit is applied on a session period which is now configured as 1 day. This means on a 24 hr window, the allocations shouldn't exceed `8.4M / 90 = 93,333` 

**Note 2**: Time periods are configured based on the relay chain block numbers and an assumption that they come around every 6 seconds.

**Note 3**: The starting block of the curve will be the relay chain block number right after the runtime upgrade. It means right after the upgrade we will be at the beginning of the fiscal period 0. However with a root permission, it's possible to set the starting block number in a way that the chain fiscal period aligns with an external fiscal period. This also allows the curve to go in a different inflation step. Obviously it's a sensitive config that's only granted to Root.

**Note 4**: When the curve starting block is configured to a be a block in the future, then the very first inflation step will be used for calculations before that time.

**Note 5**:  Regardless of how we configure the inflation steps, the limiter wouldn't allow the total issuance to exceed 21B coins.

**Note 6**: The supply inflation rate after the configured steps would simply follow the last step or lower when approaching the 21B maximum supply. which with the current configuration means, we'd approach the 21B maximum supply with a rate equal or lower than 0.1%.


pallets/allocations/src/lib.rs
```diff
@@ -107,6 +183,28 @@ pub mod pallet {
...
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		// TODO benchmark on_initialize
+		fn on_initialize(n: BlockNumberFor<T>) -> Weight {
+			let mut writes = 0;
+			let forced = <NextSessionQuota<T>>::get().is_none();
+			if T::MintCurve::get()
+				.checked_calc_next_session_quota(n, T::Currency::total_issuance(), forced)
+				.and_then(|session_quota| {
+					<NextSessionQuota<T>>::put(session_quota);
```
ETeissonniere, line 193: Shall we add an event here?

pallets/allocations/src/lib.rs
```diff
@@ -107,6 +183,28 @@ pub mod pallet {
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		// TODO benchmark on_initialize
+		fn on_initialize(n: BlockNumberFor<T>) -> Weight {
```
ETeissonniere, line 187: Do we need to run this code at the beginning of every block, or is there an optimization possible here?

Also, what's the reason between using on_initialize VS on_finalize? Are there any special cases we should be aware of?

aliXsed: The overhead on every block is very trivial. I have already optimized it to return quickly if nothing needs to be done. For our case it wouldn't matter whether we top up the quota on_initialize or on_finalize.

pallets/allocations/src/lib.rs
aliXsed: Done

pallets/allocations/src/lib.rs
```diff
@@ -107,6 +183,28 @@ pub mod pallet {
...
+				.and_then(|session_quota| {
+					<NextSessionQuota<T>>::put(session_quota);
+					Some(())
+				})
+				.is_some()
+			{
+				writes += 1;
+			}
+			if T::MintCurve::get().should_update_session_quota(n) || forced {
+				<SessionQuota<T>>::put(<NextSessionQuota<T>>::get().unwrap_or_else(Zero::zero));
```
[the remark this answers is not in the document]
aliXsed: Done

pallets/allocations/src/lib.rs
```diff
@@ -65,6 +65,74 @@ impl Default for Releases {
...
+				.unique_saturated_into();
+			let max_inflation_rate = *self
+				.inflation_steps
+				.get(step)
+				.or_else(|| self.inflation_steps.last())
+				.unwrap_or(&Zero::zero());
+			let target_increase =
+				(self.maximum_supply.saturating_sub(current_supply)).min(max_inflation_rate * current_supply);
+			let session_quota = Perbill::from_rational(self.session_period, self.fiscal_period) * target_increase;
+			Some(session_quota)
```
simonsso, line 112: There are a few unchecked multiplications

pallets/allocations/src/benchmarking.rs
```diff
@@ -22,27 +22,37 @@
...
-		let _ = ret.try_push((account, T::ExistentialDeposit::get() * 10u32.into()));
+		let _ = ret.try_push((account, T::ExistentialDeposit::get() * ALLOC_FACTOR.into()));
 	}
 	ret
 }
 
+fn assert_last_event<T: Config>(generic_event: <T as Config>::Event) {
+	let events = frame_system::Pallet::<T>::events();
+	let system_event: <T as frame_system::Config>::Event = generic_event.into();
+	let frame_system::EventRecord { event, .. } = &events[events.len() - 1];
```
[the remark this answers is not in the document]
aliXsed: It can and it would lead to an assertion which is actually the required behaviour of this function.  Also this function is taken from the other parity pallets and used exactly the same way in upstream.

pallets/allocations/src/lib.rs
aliXsed: The type Perbill features an internal saturation promise which makes it safe to be used in that way. The following passing test may better explain it:
```
#[test]
fn perbill_mul_test() {
	assert_eq!(Perbill::from_parts(u32::MAX), Perbill::from_percent(100));
	assert_eq!(Perbill::from_perthousand(u32::MAX) * u64::MAX, u64::MAX);
}
```

pallets/allocations/src/benchmarking.rs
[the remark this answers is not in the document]
simonsso: if you add assert(events.len()=>1 ) it will be clearer this it the intended behavior

pallets/allocations/src/lib.rs
```diff
@@ -102,6 +171,30 @@ pub mod pallet {
 	#[pallet::generate_store(pub(super) trait Store)]
 	pub struct Pallet<T>(PhantomData<T>);
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(n: BlockNumberFor<T>) -> Weight {
```
ETeissonniere, line 176: Shall we do like the grants/vesting pallet and use blocks from the relay chain which might be slightly more stable?

ETeissonniere (commented):
One minor comment. I am also curious as to when would the fiscal period start (date wise). Would it make sense to make it follow the fiscal quarters? If so, we might need to enforce it when deploying.

pallets/allocations/src/lib.rs
aliXsed: @ETeissonniere Switching to using the relay chain block number comes with its cons and pros. The pros as you mentioned is that the timing would be more stable. There are however a few cons. First is we couple an action on the parachain to an event on the relay chain which is first an additional reliance and coupling to the relay chain. Second you'd always need to remember to find out and use the block number from the relay chain when you need to verify or predict an action. I think we decided to pay this price for our grants because they are based on or resemble financial contracts so they'd better be more predictable time wise. However for setting an upper bound (especially that we are not going to go for a very tight bound), the simplicity and independence might be better.  What do you think?

aliXsed:
> One minor comment. I am also curious as to when would the fiscal period start (date wise). Would it make sense to make it follow the fiscal quarters? If so, we might need to enforce it when deploying.

@ETeissonniere the very first fiscal period starts at block number 0. With the current configuration it would actually follow the fiscal quarters (takes about 90 days or 648000 blocks). The first time we'd see the event `SessionQuotaCalculated` would be in the block right after the upgrade (because it's enforced for the case there hasn't been any previous calculations). Then the next time we will see that would be in the block 1296000.

pallets/allocations/src/benchmarking.rs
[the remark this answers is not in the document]
aliXsed: I added an assert for further clarification.

ETeissonniere:
@alixsed so I think we are looking at about a month difference from a chain fiscal quarter VS a company fiscal quarter. Correct?

pallets/allocations/src/lib.rs
ETeissonniere: The independence argument is certainly pretty strong. Ultimately it comes down to the guarantees we seek to provide. Do we want a very stable measure of time or are we fine with tolerating drifts of a few seconds for some blocks (which could also accumulate over time)? From what you are saying, it sounds like we could go with the former. Correct?

aliXsed: Correct, implementation wise we can go with the relay chain block pace for timing and it should be a pretty trivial change. I was on the side that a few seconds drifts and its accumulating effect for an upper bound could be okay. Especially that we know it's not going to to make intervals shorter. Parachain days will be only longer than the actual day. But this would also mean the widening or tightening of the funnel would potentially take longer than an actual quarter to happen (unless block production on parachain remains more stable going forward). I'm open to make the change and use the relay chain block pace, what would you suggest?

aliXsed:
> @aliXsed so I think we are looking at about a month difference from a chain fiscal quarter VS a company fiscal quarter. Correct?

@ETeissonniere correct. Now I see your point. Do you want a root level extrinsic to bring the start of a quarter forward?

ETeissonniere:
@aliXsed yeah, unless we can make it follow our own quarters as we deploy the upgrade

pallets/allocations/src/lib.rs
aliXsed: For the record here, the implementation is changed to use the relay chain block numbers instead.

aliXsed:
> @aliXsed yeah, unless we can make it follow our own quarters as we deploy the upgrade

@ETeissonniere I have added a root level extrinsic that allows the root to set the curve starting block number. We should however remember that for Eden the relay chain block number should be used for this purpose.

ETeissonniere:
@aliXsed after checking the code I am wondering if it wouldn't be cleaner to simply set it during the migration that would be called when we deploy the upgrade?

aliXsed:
> @aliXsed after checking the code I am wondering if it wouldn't be cleaner to simply set it during the migration that would be called when we deploy the upgrade?

@ETeissonniere there is a warning in the doc for `on_runtime_upgrade` which says:

> This function will be called before we initialized any runtime state, aka on_initialize wasn't called yet. So, information like the block number and any other block local data are not accessible.

We need to know the block number to set the curve starting block. That made me uncomfortable to take that approach. I think we can do some optimisations later without breaking anything.

ETeissonniere (approved):

PR #672, opened 2022-09-23
fix: allocations access to the relay chain block number

aliXsed:
Move quota calculations to happen in the first successful `batch` call in a fiscal period. Same for quota renewal that would happen in the first successful call in a session period. 
This new approach relies on adding two new storage for storing the schedules for quota calculations and renewals. The additional cost to most of the batch calls would be only three additional storage reads. However in those batch calls where the calculation and or renewal of quota is needed the cost would be more. 
The benefit of this new approach is that it doesn't rely on any block initialisation or finalisation process. And if there is no batch call in a block, there is now overhead related to the allocations for the block. The implementation is also capable of being based on both relay chain block numbers or parachain block numbers and doesn't need to happen in all the blocks.
The disadvantage is first a batch call occasionally does stuff not directly related to allocating rewards to the given list. Also when a batch call fails even though the call is not decorated with `#[transactional]` macro, all the storage changes including the calculation or renewal of the quota is reverted which leads to the same calculations and updates to be needed for the next call.

### DIFF
--- a/pallets/allocations/src/benchmarking.rs
+++ b/pallets/allocations/src/benchmarking.rs
@@ -60,7 +60,6 @@ benchmarks! {
 		let mut members = <BenchmarkOracles<T>>::get();
 		assert!(members.try_push(oracle.clone()).is_ok());
 		<BenchmarkOracles<T>>::put(&members);
-		<Allocations<T>>::checked_update_session_quota();
 		<SessionQuota<T>>::put(T::ExistentialDeposit::get() * (b * ALLOC_FACTOR).into());
 	}:{
 		let _ = Allocations::<T>::allocate(batch_arg);
@@ -80,6 +79,14 @@ benchmarks! {
 	}
 	verify {
 		assert_last_event::<T>(Event::SessionQuotaRenewed.into())
+	}
+
+	checked_update_session_quota {
+	}:{
+		Allocations::<T>::checked_update_session_quota();
+	}
+	verify {
+		assert_eq!(frame_system::Pallet::<T>::events().len(), 2);
 	}
 
 	set_curve_starting_block {
--- a/pallets/allocations/src/benchmarking.rs
+++ b/pallets/allocations/src/benchmarking.rs
@@ -60,20 +60,21 @@ benchmarks! {
 		let mut members = <BenchmarkOracles<T>>::get();
 		assert!(members.try_push(oracle.clone()).is_ok());
 		<BenchmarkOracles<T>>::put(&members);
+		<Allocations<T>>::checked_update_session_quota();
 		<SessionQuota<T>>::put(T::ExistentialDeposit::get() * (b * ALLOC_FACTOR).into());
 	}: _(RawOrigin::Signed(oracle), batch_arg)
 
 	calc_quota {
 	}: {
-		Allocations::<T>::checked_calc_session_quota(Zero::zero(), true);
+		Allocations::<T>::checked_calc_session_quota(Zero::zero());
 	}
 	verify {
-		assert_last_event::<T>(Event::SessionQuotaCalculated(<NextSessionQuota<T>>::get().unwrap()).into())
+		assert_last_event::<T>(Event::SessionQuotaCalculated(<NextSessionQuota<T>>::get()).into())
 	}
 
 	renew_quota {
 	}: {
-		Allocations::<T>::checked_renew_session_quota(Zero::zero(), true);
+		Allocations::<T>::checked_renew_session_quota(Zero::zero());
 	}
 	verify {
 		assert_last_event::<T>(Event::SessionQuotaRenewed.into())
--- a/pallets/allocations/src/benchmarking.rs
+++ b/pallets/allocations/src/benchmarking.rs
@@ -52,7 +52,7 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::Event) {
 }
 
 benchmarks! {
-	batch {
+	allocate {
 		let b in 1..T::MaxAllocs::get();
 
 		let batch_arg = make_batch::<T>(b);
@@ -62,7 +62,9 @@ benchmarks! {
 		<BenchmarkOracles<T>>::put(&members);
 		<Allocations<T>>::checked_update_session_quota();
 		<SessionQuota<T>>::put(T::ExistentialDeposit::get() * (b * ALLOC_FACTOR).into());
-	}: _(RawOrigin::Signed(oracle), batch_arg)
+	}:{
+		let _ = Allocations::<T>::allocate(batch_arg);
+	}
 
 	calc_quota {
 	}: {
--- a/pallets/allocations/src/lib.rs
+++ b/pallets/allocations/src/lib.rs
@@ -33,9 +33,9 @@ use frame_support::{
 
 use frame_system::ensure_signed;
 use scale_info::TypeInfo;
-use sp_arithmetic::traits::{CheckedRem, UniqueSaturatedInto};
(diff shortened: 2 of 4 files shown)


pallets/allocations/src/lib.rs
```diff
@@ -82,57 +81,83 @@ impl<T: Config> MintCurve<T> {
...
+		} else {
+			let s = curve_start.saturating_sub(
+				curve_start
+					.saturating_sub(n)
+					.checked_div(&period)
+					.unwrap_or_else(Bounded::max_value)
+					.saturating_mul(period),
+			);
+			if n == s {
+				n.saturating_add(period)
```
simonsso, line 155: will this not drift if call is delayed?

pallets/allocations/src/lib.rs
```diff
@@ -333,42 +361,57 @@ impl<T: Config> Pallet<T> {
...
-	fn checked_calc_session_quota(n: T::BlockNumber, forced: bool) -> Weight {
-		if let Some(session_quota) =
-			T::MintCurve::get().checked_calc_next_session_quota(n, T::Currency::total_issuance(), forced)
-		{
-			<NextSessionQuota<T>>::put(session_quota);
-			Self::deposit_event(Event::SessionQuotaCalculated(session_quota));
-			T::WeightInfo::calc_quota()
-		} else {
-			T::DbWeight::get().reads(1 as Weight) // Storage: Balances TotalIssuance (r:1 w:0)
+	fn checked_update_session_quota() {
```
simonsso, line 364: Slightly longer functions would improve readability, it is not obvious which are pure functions and where there are side effects.

pallets/allocations/src/lib.rs
```diff
@@ -82,57 +81,83 @@ impl<T: Config> MintCurve<T> {
...
+	fn next_schedule(n: T::BlockNumber, curve_start: T::BlockNumber, period: T::BlockNumber) -> T::BlockNumber {
+		if n >= curve_start {
+			n.saturating_sub(curve_start)
+				.checked_div(&period)
+				.unwrap_or_else(Bounded::max_value)
+				.saturating_add(One::one())
+				.saturating_mul(period)
+				.saturating_add(curve_start)
+		} else {
+			let s = curve_start.saturating_sub(
```
simonsso, line 147: `s` could have a more descriptive name

pallets/allocations/src/lib.rs
aliXsed: All the pallet functions have access to the pallet storage and regardless of their look they can not be considered pure functions, it's substrate design. Only the MintCurve functions don't have access to the pallet storage and there is only one pure function called MintCurve::next_schedule.

pallets/allocations/src/lib.rs
aliXsed: An extrinsic is executed in an atomic way, meaning there wouldn't be a drift in the block number or an execution of any other extrinsic in parallel. This should be through for the relay chain block number as well, because every single call should be executed by parachain validators to be backed in a certain relay chain block.

ETeissonniere:
> Also when a batch call fails even though the call is not decorated with #[transactional] macro, all the storage changes including the calculation or renewal of the quota is reverted which leads to the same calculations and updates to be needed for the next call.
@aliXsed shall we make it explicit anyways?

pallets/allocations/src/lib.rs
```diff
@@ -180,88 +207,35 @@ pub mod pallet {
...
-		}
-	}
-
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		/// Optimized allocation call, which will batch allocations of various amounts
 		/// and destinations and together. This allow us to be much more efficient and thus
 		/// increase our chain's capacity in handling these transactions.
-		#[pallet::weight(<T as pallet::Config>::WeightInfo::batch(batch.len().try_into().unwrap_or_else(|_| T::MaxAllocs::get())))]
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::allocate(rewards.len().try_into().unwrap_or_else(|_| T::MaxAllocs::get())))]
```
ETeissonniere, line 215: why are we renaming the weight function?

pallets/allocations/src/benchmarking.rs
```diff
@@ -52,28 +52,31 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::Event) {
...
 		let oracle: T::AccountId = account("oracle", 0, SEED);
 		let mut members = <BenchmarkOracles<T>>::get();
 		assert!(members.try_push(oracle.clone()).is_ok());
 		<BenchmarkOracles<T>>::put(&members);
+		<Allocations<T>>::checked_update_session_quota();
 		<SessionQuota<T>>::put(T::ExistentialDeposit::get() * (b * ALLOC_FACTOR).into());
-	}: _(RawOrigin::Signed(oracle), batch_arg)
+	}:{
+		let _ = Allocations::<T>::allocate(batch_arg);
+	}
```
ETeissonniere, line 67: We should benchmark for the worst case scenario, looks like line 63 would let the benchmarked extrinsic skip it next time it runs based on my understanding of `checked_update_session_quota`

pallets/allocations/src/lib.rs
ETeissonniere: Additionally, this does not seem to represent the worst case scenario based on line 225

ETeissonniere (changes requested):

pallets/allocations/src/benchmarking.rs
aliXsed: @ETeissonniere Actually the line 63 is not needed any more is a remnant of the previous commits. The function `allocate` doesn't call checked_update_session_quota. I made this separation intentionally to be able to calculate the weights accurately.

pallets/allocations/src/lib.rs
aliXsed: Good catch, we'll fix it soon

aliXsed:
> > Also when a batch call fails even though the call is not decorated with #[transactional] macro, all the storage changes including the calculation or renewal of the quota is reverted which leads to the same calculations and updates to be needed for the next call.
> 
> @aliXsed shall we make it explicit anyways?

@ETeissonniere because we don't really need it to be transactional, I would say leave it the way it is.

ETeissonniere (dismissed):

pallets/allocations/src/lib.rs
```diff
@@ -180,88 +207,35 @@ pub mod pallet {
...
 	impl<T: Config> Pallet<T> {
 		/// Optimized allocation call, which will batch allocations of various amounts
 		/// and destinations and together. This allow us to be much more efficient and thus
 		/// increase our chain's capacity in handling these transactions.
-		#[pallet::weight(<T as pallet::Config>::WeightInfo::batch(batch.len().try_into().unwrap_or_else(|_| T::MaxAllocs::get())))]
+		#[pallet::weight(T::WeightInfo::allocate(rewards.len().try_into().unwrap_or_else(|_| T::MaxAllocs::get())).saturating_add(T::WeightInfo::checked_update_session_quota()))]
 		pub fn batch(
 			origin: OriginFor<T>,
-			batch: BoundedVec<(T::AccountId, BalanceOf<T>), T::MaxAllocs>,
+			rewards: BoundedVec<(T::AccountId, BalanceOf<T>), T::MaxAllocs>,
```
simonsso, line 218: It is a breaking change for to change this parameter name. It will require a coordinated deployment on the testnet

simonsso (approved):

PR #673, opened 2022-09-29
Upgrade to polkadot v0.9.29

simonsso:
- [x]  Verify merges
- [x]  Upgrade spec_version in runtimes/eden/src/version.rs
- [x]  cargo build --release
- [x]  cargo test
- [x]  Benchmarks compiles
- [x]  Weights update or ignored
- [x]  try-runtime
- [x]  Run local test-net
- [ ]  File tickets if metadata changed
- [x]  Clean up debug commits


pallets/allocations/src/weights.rs
```diff
@@ -67,13 +67,13 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	/// The range of component `b` is `[1, 500]`.
 	fn allocate(b: u32) -> Weight {
-		(0 as Weight)
-			// Standard Error: 4_000
-			.saturating_add((7_946_000 as Weight).saturating_mul(b as Weight))
-			.saturating_add(T::DbWeight::get().reads(8 as Weight))
-			.saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(b as Weight)))
-			.saturating_add(T::DbWeight::get().writes(6 as Weight))
-			.saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(b as Weight)))
+		Weight::from_ref_time(0 as u64)
```
aliXsed, line 70: You can remove all the `as u64` for integral literals in this file as suggested by clippy as well.

aliXsed: Actually I remembered this is a generated code and we better not touch it.

pallets/grants/src/lib.rs
```diff
@@ -34,7 +34,7 @@ use frame_support::{
 	BoundedVec,
 };
 use sp_runtime::{
-	traits::{AtLeast32Bit, BlockNumberProvider, CheckedAdd, Saturating, StaticLookup, Zero},
+	traits::{AtLeast32Bit, BlockNumberProvider, CheckedAdd, StaticLookup, Zero, Saturating},
```
aliXsed, line 37: Have a formatted the code?

pallets/allocations/src/weights.rs
aliXsed: Actually never mind this. I re-benchmarked the allocations and the generated code would look like that.

aliXsed (dismissed):
LGMT!

noisefloors:
Looks good to me. In my opinion, we should consider updating the weights template to remove the unnecessary cast that Clippy is complaining about, but that's for another time.

noisefloors (approved):

PR #679, opened 2022-10-31
Replace POA with Collator selection

aliXsed:
Must:
- [x] Add pallet_collator_selection
- [x] Remove pallet_poa
- [x] Set Nodle collators as invulnerables in pallet_collator_selction. This is so the Nodle collators stay selected the chain's governance votes for removing them. The down side is when Nodle collators are unstable and not creating blocks, they wouldn't be kicked out. As a result the missed turns continue to show up as repeating extra delays in block creation. Fortunately, this hasn't been an issue with Nodle's main-net is only observed on the test-net.
- [x] Pay the 80% share of the fees that now goes to the block author to the stake pot held by pallet-collator-selection so this pallet compensates the authors based on its own logic.
- [x] Fix try-runtime panicked at 'Empty validator set for session 0 in genesis block!', 
- [x] Migrate ValidatorsSet storage, Remove Poa strorage
- [x] Test the change locally
- [x] Test an upgrade from the current runtime to this new runtime locally and check the block creation proceeds with the previous set of the collators.
- [ ] Update weights


runtimes/eden/src/lib.rs
```diff
@@ -93,13 +94,13 @@ construct_runtime! {
 		TechnicalMembership: pallet_membership::<Instance3> = 17,
 
 		// Consensus
+		CollatorSelection: pallet_collator_selection::{Pallet, Call, Storage, Event<T>, Config<T>} = 19,
```
ETeissonniere, line 97: Can we use `26`? I was trying to keep the `1*` prefix for governance and `2*` for consensus

ETeissonniere: In hindsight we could also reuse `21` or `22`

runtimes/eden/src/migrations.rs
```diff
@@ -0,0 +1,77 @@
...
+	pub const MaxMembers: u32 = 50;
+}
+
+pub struct MoveValidatorsSetToInvulnerables;
+impl OnRuntimeUpgrade for MoveValidatorsSetToInvulnerables {
+	fn on_runtime_upgrade() -> Weight {
+		if let Some(validators) =
+			migration::take_storage_value::<BoundedVec<AccountId, MaxMembers>>(b"ValidatorsSet", b"Members", &[])
+		{
+			log::info!(target: "try-runtime", "ValidatorsSet::Members are {:?}", validators);
```
ETeissonniere, line 21:
```suggestion
			log::info!(target: "migration", "ValidatorsSet::Members are {:?}", validators);
```
Technically this isn't run only in the `try-runtime` target but also on a real migration

runtimes/eden/src/migrations.rs
```diff
@@ -0,0 +1,77 @@
...
+			} else {
+				log::error!(target: "try-runtime", "Failed to remove ValidatorsSet::Prime completely");
+			}
+
+			let clear_poa_result = migration::clear_storage_prefix(b"Poa", b"StorageVersion", &[], None, None);
+			if clear_poa_result.maybe_cursor.is_none() {
+				log::info!(target: "try-runtime", "Poa::StorageVersion with {} unique entries is removed", clear_poa_result.unique);
+			} else {
+				log::error!(target: "try-runtime", "Failed to remove Poa::StorageVersion completely");
+			}
```
ETeissonniere, line 37: I think you must have a way to clear all the storage values under a certain prefix to make this a little less verbose

runtimes/eden/src/migrations.rs
```diff
@@ -0,0 +1,77 @@
...
+// MaxMembers is chosen based on what used to be the MaxMembers param for the pallet ValidaorsSet
+// We have intentionally used the same number for MaxInvulnerables for the pallet CollatorSelection
+// So the BoundedVec storage in these two pallets remain compatible.
+parameter_types! {
+	pub const MaxMembers: u32 = 50;
+}
+
+pub struct MoveValidatorsSetToInvulnerables;
+impl OnRuntimeUpgrade for MoveValidatorsSetToInvulnerables {
+	fn on_runtime_upgrade() -> Weight {
```
ETeissonniere, line 17: Shall we configure the candidacy bond too?

runtimes/eden/src/lib.rs
aliXsed: I wanted to be very cautious to not reuse pallet ids on subsequent upgrades. This may trigger very weird bugs/issues in the tools which explore the blocks but don't update the metadata in the right place. Though it should be the concern of those tools, but I wanted not to leave any room for such baffling/misleading issues.

runtimes/eden/src/migrations.rs
aliXsed: I couldn't find one in those migration helper functions. They all need both the module and the item to be specified. I tried to see if an empty item name can do the job but it didn't. Do you know how we can do that?

runtimes/eden/src/migrations.rs
aliXsed: We can if we have a decision on how much it should be. We can also do that after the upgrade. What do you suggest?

runtimes/eden/src/lib.rs
ETeissonniere: Makes sense, shall we do `26` then?

runtimes/eden/src/migrations.rs
ETeissonniere: Check this one https://paritytech.github.io/substrate/master/frame_support/storage/migration/fn.clear_storage_prefix.html

runtimes/eden/src/migrations.rs
ETeissonniere: Can you set it to 3M NODL? I think that's what was floating around internally for now

runtimes/eden/src/lib.rs
aliXsed: It's necessary for CollatorSelection to be declared before Session. I can still keep it before Session and give it the number 26 but that wouldn't look good and may tempt people to reorder it which will be breaking.

runtimes/eden/src/migrations.rs
aliXsed: I'm using the exact same function, but there you need to specify the "item". I don't know how we can remove the while storage a pallet in a single call.

runtimes/eden/src/migrations.rs
aliXsed: Done

ETeissonniere (approved):

PR #680, opened 2022-11-08
Add smart contracts

aliXsed:
Bring pallet_contracts to Eden runtime with the safest options.


runtimes/eden/Cargo.toml
```diff
@@ -122,12 +126,19 @@ try-runtime = [
...
   "cumulus-pallet-parachain-system/try-runtime",
   "cumulus-pallet-xcm/try-runtime",
   "parachain-info/try-runtime",
 ]
+# Make contract callable functions marked as __unstable__ available. Do not enable
+# on live chains as those are subject to change.
+contracts-unstable-interface = [
+  "pallet-contracts/unstable-interface"
+]
+
```
ETeissonniere, line 141: Soooo shall we or shall we not enable this?

runtimes/eden/src/lib.rs
```diff
@@ -122,6 +119,9 @@ construct_runtime! {
 
 		// DAO
 		DaoReserve: pallet_reserve::<Instance4> = 60,
+
+		// Smart Contracts.
+		Contracts: pallet_contracts::{Pallet, Call, Storage, Event<T>} = 62,
```
ETeissonniere, line 124:
```suggestion
		Contracts: pallet_contracts = 62,
```
I don't think you need the syntactic sugar anymore.

runtimes/eden/src/pallets_util.rs
```diff
@@ -138,3 +139,47 @@ impl pallet_uniques::Config for Runtime {
 	type CreateOrigin = AsEnsureOriginWithArg<EnsureSigned<AccountId>>;
 	type Locker = ();
 }
+
+parameter_types! {
+	pub const DepositPerItem: Balance = constants::deposit(1, 0);
+	pub const DepositPerByte: Balance = constants::deposit(0, 1);
```
ETeissonniere, line 145: So what is this in NODL?

runtimes/eden/Cargo.toml
aliXsed: We would need to enable this only for local tests

runtimes/eden/src/pallets_util.rs
aliXsed: It goes to our pre-existing function deposit where we have had specified how much per item and per byte should be deposited:
```
pub const fn deposit(items: u32, bytes: u32) -> Balance {
	items as Balance * 1_500 * MICRO_NODL + (bytes as Balance) * 600 * MICRO_NODL
}
```

ETeissonniere: So `DepositPerItem` is `1_500 * MICRO_NODL` and `DepositPerByte` is `600 * MICRO_NODL`. Any reasons for not making it explicit?

aliXsed: The `deposit` function is reused for configuring pallet_multisig and pallet_preimage and potentially more in the future. It's for the sake of consistency that we follow the same pricing scheme everywhere.

runtimes/eden/Cargo.toml
ETeissonniere: Can this be set to tests only?

runtimes/eden/src/pallets_util.rs
ETeissonniere: Makes sense

runtimes/eden/Cargo.toml
aliXsed: It's not for unit tests, but this feature should only be enabled manually while building the runtime. So as long as it's not appearing in our build instructions which is the case now, it's not doing anything.

ETeissonniere: I see, got it

ETeissonniere (approved):

PR #685, opened 2022-12-15
Rework CI

simonsso:


runtimes/eden/src/constants.rs
```diff
@@ -124,4 +124,12 @@ mod tests {
 		assert_eq!(100 * MICRO_NODL, CENTS);
 		assert_eq!(100 * NANO_NODL, MILLICENTS);
 	}
+
+	#[test]
+	fn test_deposit() {
```
aliXsed, line 129: Thanks for adding this.

.github/workflows/ci.yml
```diff
@@ -86,9 +105,9 @@ jobs:
         with:
           fail_ci_if_error: false
 
-
   try-runtime:
     runs-on: ubuntu-latest-16-cores
+    timeout-minutes: 45
```
aliXsed, line 110: What happens when it times out? Is it a failure?

simonsso: Yes it will show as ❌ in the summary, but when one open it will be a warning notification
image

Since we push several times when developing this way it is easy to see if it was just once it got stuck in network traffic or it is a failing trend.

runtimes/eden/src/constants.rs
simonsso: It was low hanging fruit ;-)

aliXsed (approved):
LGTM

PR #688, opened 2022-12-19
Add cache for try-runtime

simonsso:


simonsso:
Number of warnings down to only 3.

.github/workflows/ci.yml
```diff
@@ -106,27 +102,43 @@ jobs:
 
   try-runtime:
     runs-on: ubuntu-latest-16-cores
-    timeout-minutes: 45
+    timeout-minutes: 90
```
simonsso, line 105:
```suggestion
    timeout-minutes: 50
```

runtimes/eden/src/pallets_consensus.rs
```diff
@@ -49,8 +49,8 @@ impl pallet_session::Config for Runtime {
 	type SessionManager = CollatorSelection;
 	type ShouldEndSession = pallet_session::PeriodicSessions<Period, Offset>;
 	type NextSessionRotation = pallet_session::PeriodicSessions<Period, Offset>;
-	type Event = Event;
-	type SessionHandler = <SessionKeys as sp_runtime::traits::OpaqueKeys>::KeyTypeIdProviders;
+		type Event = Event;
+		type SessionHandler = <SessionKeys as sp_runtime::traits::OpaqueKeys>::KeyTypeIdProviders;
```
simonsso, line 53:
```suggestion
	type Event = Event;
	type SessionHandler = <SessionKeys as sp_runtime::traits::OpaqueKeys>::KeyTypeIdProviders;
```

.github/workflows/ci.yml
```diff
@@ -14,18 +14,43 @@ env:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ env.nightly }}
+          target: ${{ env.target }}
+          components: rustfmt
+
+      - name: Rustfmt Check
+        uses: actions-rust-lang/rustfmt@v1
```
aliXsed, line 43: I think it could be possibly more efficient for this part to be a step for the job "clippy". Though in that case you could call it "fmt-clippy".

.github/workflows/ci.yml
```diff
@@ -51,24 +76,11 @@ jobs:
...
-          command: fmt
-
-      - name: Clippy
-        uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features -- -A clippy::type_complexity -A clippy::identity_op -A clippy::boxed_local -D dead_code
-          toolchain: ${{ env.nightly }}
-
       - name: Build
```
aliXsed, line 79: Contrary to my previous comment, here, since build and test steps are not helping each other, they could be separate jobs.

simonsso: I was confused by the cache step here and did not want to touch it. But yes it will probably be split

.github/workflows/ci.yml
aliXsed: If the overhead is trivial, then up to you. However if you want to keep them as steps but let them both run even when one is failed, you can use `if: success() || failure()` as suggested [here](https://stackoverflow.com/a/58859404/2746280)

.github/workflows/ci.yml
aliXsed: I see your point. I think it would be good to have the unit test as a separate job since the build for tests is totally different from the normal build.

.github/workflows/ci.yml
```diff
@@ -106,27 +113,43 @@ jobs:
...
-          target: ${{ env.target }}
+      - name: Check Version
+        run: |
+         echo -n "eden_rev=" >> $GITHUB_OUTPUT
+         curl -s --request POST   --url https://nodle-parachain.api.onfinality.io/public   --header 'Content-Type: application/json'   --data '{
+             "jsonrpc": "2.0", 
+              "method": "system_version",    
+              "params": [],
+              "id": 1
+            }' | jq '{"result"}[]' >> $GITHUB_OUTPUT
```
aliXsed, line 127: Nice!

.github/workflows/ci.yml
```diff
@@ -106,27 +113,43 @@ jobs:
...
-      - name: Install latest stable Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          target: ${{ env.target }}
+      - name: Check Version
+        run: |
+         echo -n "eden_rev=" >> $GITHUB_OUTPUT
+         curl -s --request POST   --url https://nodle-parachain.api.onfinality.io/public   --header 'Content-Type: application/json'   --data '{
```
aliXsed, line 122:
```suggestion
         curl -s --request POST  --url ${{ env.try-runtime-uri }} --header 'Content-Type: application/json'  --data '{
```

.github/workflows/ci.yml
```diff
@@ -106,27 +113,43 @@ jobs:
...
+        if: steps.cachedir.outputs.cache-hit != 'true'
+        run: |
+            install -d snapshots
+            date > snapshots/created_at
+            cargo run --release --bin nodle-parachain --features=try-runtime try-runtime --execution native --chain ${{ env.try-runtime-chain }} --no-spec-name-check  on-runtime-upgrade  live --snapshot-path="snapshots/paradis-snapshot-full" -u ${{ env.try-runtime-uri}}
+      - name: Try runtime reuse snap
+        if: steps.cachedir.outputs.cache-hit == 'true'
+        run: |
+            cat snapshots/created_at
+            cargo run --release --bin nodle-parachain --features=try-runtime try-runtime --execution native --chain ${{ env.try-runtime-chain }} --no-spec-name-check  on-runtime-upgrade  snap --snapshot-path="snapshots/paradis-snapshot-full"
```
aliXsed, line 153: Good job! man

.github/workflows/ci.yml
simonsso: They are not the same:  `try-runtime-uri: wss://eden-rpc.dwellir.com:443` one is https and the other is wss (which is https)

node/src/chain_spec.rs
```diff
@@ -16,8 +16,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
aliXsed, line 16: The standard format is a bit different from the nightly format that we used to check in CI. You may need to either reformat your workspace (if the change is not huge) or go back to using the nightly in CI

.github/workflows/ci.yml
aliXsed: Ah ok I missed to see. Would it be good to define this uri in env as well?

node/src/chain_spec.rs
simonsso: This is an error I inserted for rustfmt to trap ;-)

.github/workflows/ci.yml
```diff
@@ -14,18 +14,41 @@ env:
...
+      - uses: actions/checkout@v3
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ env.nightly }}
+          target: ${{ env.target }}
+          components: clippy
+      - name: Clippy
+        run: |
+           unset RUSTFLAGS
+           cargo clippy --all-features -- -W clippy::type_complexity -W clippy::identity_op -W clippy::boxed_local -D dead_code
```
simonsso, line 29: `cargo clippy` did not work as well as the old addon. (warnings are not visible on the code review screen) Will look for a maintained fork.

runtimes/eden/src/lib.rs
```diff
@@ -90,7 +90,7 @@ construct_runtime! {
 		TechnicalCommittee: pallet_collective::<Instance1> = 16,
 		TechnicalMembership: pallet_membership::<Instance3> = 17,
 
-		// Consensus
+	// Consensus
```
simonsso, line 93:
```suggestion
		// Consensus
```

PR #696, opened 2023-01-18
[XCM] Integration of the pallet asset for foreign asset management

Hounsette:
- [ ] Pallet configuration
- [ ] Asset can be only created through a mandate
- [ ] Asset can be only minte through xcm instruction


Hounsette:
Not relevant anymore

PR #707, opened 2023-02-20
Replace actions-rs/clippy with maintained fork

simonsso:
actions-rs gh actions is unmaintained this is a fork which is slightly more maintained


simonsso:
Same fix as in chain-minions

aliXsed (approved):

PR #712, opened 2023-02-27
Run all CI on 8 cores

simonsso:


simonsso:
@aliXsed  I think we should change the executors to 8 cores for cost savings. Executors are prices linear and performance gains are always under 100% for doubling the cores.

However at 8 core all builds are under 1 hour.

.github/workflows/ci.yml
```diff
@@ -90,7 +90,7 @@ jobs:
             output.txt
 
   tarpaulin:
-    runs-on: ubuntu-latest-16-cores
+    runs-on: ubuntu-latest-8-cores
```
simonsso, line 93: This could run fine on 4 cores too.

aliXsed: let's do that

.github/workflows/ci.yml
```diff
@@ -89,7 +89,7 @@ jobs:
...
```
simonsso, line 92:
```suggestion
    runs-on: ubuntu-latest-4-cores
```

simonsso (commented):
Looks good for me now

aliXsed (approved):

PR #715, opened 2023-03-07
Increase max nft string length

simonsso:
If the space allowed for a few more bytes IPFS:// urls could for instance be used.

### DIFF
--- a/runtimes/eden/src/pallets_util.rs
+++ b/runtimes/eden/src/pallets_util.rs
@@ -113,7 +113,7 @@ parameter_types! {
 	pub const MetadataDepositPerByte: Balance = 10 * constants::MILLI_NODL;
 	pub const KeyLimit: u32 = 32;
 	pub const ValueLimit: u32 = 256;
-	pub const StringLimit: u32 = 50;
+	pub const StringLimit: u32 = 64;
 }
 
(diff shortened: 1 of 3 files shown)


aliXsed (approved):
@simonsso I noticed for system parachains such as statemint they have gone all the way to "128" for this limit. So if it will be better future proof to go for 128 it would be still approved from my side.

simonsso:
> @simonsso I noticed for system parachains such as statemint they have gone all the way to "128" for this limit. So if it will be better future proof to go for 128 it would be still approved from my side.

Let's do that.

runtimes/eden/src/pallets_util.rs
```diff
@@ -113,7 +113,7 @@ parameter_types! {
 	pub const MetadataDepositPerByte: Balance = 10 * constants::MILLI_NODL;
 	pub const KeyLimit: u32 = 32;
 	pub const ValueLimit: u32 = 256;
-	pub const StringLimit: u32 = 50;
+	pub const StringLimit: u32 = 64;
```
simonsso, line 116:
```suggestion
	pub const StringLimit: u32 = 128;
```

aliXsed (approved):

PR #723, opened 2023-03-24
correct storage of new pallets and prepare release

aliXsed:
The on chain version of the pallets should match the pallets versions for newly added pallets. Also the safe xcm version for PolkadotXcm should be set manually.

### DIFF
--- a/runtimes/eden/src/migrations.rs
+++ b/runtimes/eden/src/migrations.rs
@@ -1,6 +1,12 @@
 use crate::Runtime;
-use frame_support::{migration, parameter_types, traits::OnRuntimeUpgrade, weights::Weight, BoundedVec};
+use frame_support::{
+	migration, parameter_types,
+	traits::{GetStorageVersion, OnRuntimeUpgrade},
+	weights::Weight,
+	BoundedVec,
+};
 use primitives::{AccountId, Balance};
+use xcm::prelude::{XcmVersion, XCM_VERSION};
 
 #[cfg(feature = "try-runtime")]
 use codec::{Decode, Encode};
@@ -33,9 +39,13 @@ const TRY_RUNTIME: &str = "try-runtime";
 
 const CANDIDACY_BOND: Balance = 3_000_000 * crate::constants::NODL;
 
+const POLKADOT_XCM_MODULE: &[u8] = b"PolkadotXcm";
+const SAFE_XCM_VERSION: &[u8] = b"SafeXcmVersion";
+
 pub struct MoveValidatorsSetToInvulnerables;
 impl OnRuntimeUpgrade for MoveValidatorsSetToInvulnerables {
 	fn on_runtime_upgrade() -> Weight {
+		let mut weight = <Runtime as frame_system::Config>::DbWeight::get().reads(5);
 		if let Some(validators) = migration::take_storage_value::<BoundedVec<AccountId, MaxMembers>>(
 			VALIDATORS_SET_MODULE,
 			MEMBERS_ITEM,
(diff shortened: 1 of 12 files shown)


runtimes/eden/src/migrations.rs
```diff
@@ -33,9 +39,13 @@ const TRY_RUNTIME: &str = "try-runtime";
 
 const CANDIDACY_BOND: Balance = 3_000_000 * crate::constants::NODL;
 
+const POLKADOT_XCM_MODULE: &[u8] = b"PolkadotXcm";
+const SAFE_XCM_VERSION: &[u8] = b"SafeXcmVersion";
+
 pub struct MoveValidatorsSetToInvulnerables;
 impl OnRuntimeUpgrade for MoveValidatorsSetToInvulnerables {
 	fn on_runtime_upgrade() -> Weight {
+		let mut weight = <Runtime as frame_system::Config>::DbWeight::get().reads(5);
```
simonsso, line 48: Make an assignment here of the full `if let` and avoid the `mut`

runtimes/eden/src/migrations.rs
```diff
@@ -75,10 +85,43 @@ impl OnRuntimeUpgrade for MoveValidatorsSetToInvulnerables {
...
+		if contracts_version != <pallet_contracts::pallet::Pallet<Runtime>>::on_chain_storage_version() {
+			contracts_version.put::<pallet_contracts::pallet::Pallet<Runtime>>();
+			weight += <Runtime as frame_system::Config>::DbWeight::get().writes(1);
+		} else {
+			log::info!(target: MIGRATION, "No migration for Contracts");
+		}
+
+		if migration::get_storage_value::<XcmVersion>(POLKADOT_XCM_MODULE, SAFE_XCM_VERSION, EMPTY_HASH).is_none() {
+			migration::put_storage_value(POLKADOT_XCM_MODULE, SAFE_XCM_VERSION, EMPTY_HASH, Some(XCM_VERSION));
+			weight += <Runtime as frame_system::Config>::DbWeight::get().writes(1);
```
simonsso, line 119: aha I see how this is structured now

runtimes/eden/src/migrations.rs
```diff
@@ -100,27 +143,63 @@ impl OnRuntimeUpgrade for MoveValidatorsSetToInvulnerables {
...
+			"CollatorSelection::Invulnerables are {:?}",
+			invulnerables
+		);
+		let candidacy_bond = pallet_collator_selection::pallet::Pallet::<Runtime>::candidacy_bond();
+		log::info!(
+			target: TRY_RUNTIME,
+			"CollatorSelection::candidacy_bond is {:?}",
+			candidacy_bond
+		);
+		let validators: BoundedVec<AccountId, MaxMembers> = Decode::decode(&mut state.as_slice()).unwrap();
```
simonsso, line 161: could this unwrap be a `?`?

runtimes/eden/src/migrations.rs
```diff
@@ -100,27 +143,63 @@ impl OnRuntimeUpgrade for MoveValidatorsSetToInvulnerables {
...
+
+		if <pallet_contracts::pallet::Pallet<Runtime>>::current_storage_version()
+			!= <pallet_contracts::pallet::Pallet<Runtime>>::on_chain_storage_version()
+		{
+			log::info!(target: TRY_RUNTIME, "Contracts storage version is not updated");
+			return Err("Contracts storage version is not updated");
+		}
+		log::info!(target: TRY_RUNTIME, "Contracts storage version was good");
+
+		if let Some(xcm_version) =
```
simonsso, line 191: Use this a a final statement and embed the Ok(()) in the correct place and end the others without return, then the compiler will check all arms are covered for us.

runtimes/eden/src/migrations.rs
aliXsed: Yes and it's actual better. Thank you for seeing it.

runtimes/eden/src/migrations.rs
aliXsed: I think because I didn't have `else` statements in my structure, I had to `return` explicitly.

PR #725, opened 2023-03-27
feat(runtime-eden): remove migrations code 

aliXsed:
since upgrade has gone through, the migration logic should be removed


ETeissonniere (approved):

01SMD (commented):
.

PR #728, opened 2023-04-06
fix(runtime-eden): recognise xtokens' nodlNative as known asset

aliXsed:
We refer to NODL as an asset represented by our pallet instance 2 (pallet-balances) on our parachain and this location should be regarded as a known asset type.
Also fix xTokens to use the same weigher as xcm -executor.

### DIFF
--- a/runtimes/eden/src/xcm_config.rs
+++ b/runtimes/eden/src/xcm_config.rs
@@ -21,9 +21,9 @@ use sp_runtime::traits::Convert;
 use xcm::{latest::NetworkId, latest::Weight as XcmWeight, prelude::*};
 use xcm_builder::{
 	AccountId32Aliases, AllowKnownQueryResponses, AllowSubscriptionsFrom, AllowTopLevelPaidExecutionFrom,
-	CurrencyAdapter, EnsureXcmOrigin, FixedWeightBounds, IsConcrete, LocationInverter, NativeAsset, ParentIsPreset,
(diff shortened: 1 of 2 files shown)


runtimes/eden/src/xcm_config.rs
```diff
@@ -206,7 +206,7 @@ impl orml_xtokens::Config for Runtime {
 	type AccountIdToMultiLocation = AccountIdToMultiLocation;
 	type SelfLocation = SelfLocation;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type Weigher = FixedWeightBounds<UnitWeightCost, RuntimeCall, MaxInstructions>;
+	type Weigher = WeightInfoBounds<crate::weights::NodleXcmWeight<RuntimeCall>, RuntimeCall, MaxInstructions>;
```
Hounsette, line 209: I suspect this weigher to not be used as I did my tests and the weigher from the pallet_xcm was making the transaction to fail

Hounsette:
LGTM, tested locally!

Hounsette (approved):

runtimes/eden/src/xcm_config.rs
aliXsed: Yes you can change pallet_xcm to use `FixedWeightBounds` or xTokens to use `WeightInfoBounds`. The point is both of them seem to need to use the same weigher. In this case, the benchmarked weigher is the xcm one and that's why I changed xTokens.

01SMD (commented):
.

PR #733, opened 2023-04-20
chore: prepare to release 2.0.30

aliXsed:
We need to have a minor release which contains the fix for our xtoken weigher.


01SMD (commented):
.

PR #735, opened 2023-04-26
Upgrade to upstream 0.9.40

aliXsed:
- [x] Fix xcm benchmarks for generic calls
- [x] Use the new weight calculating functions from xcm benchmarks
- [x] Benchmark the PR on the standard machine and update the weights


node/src/chain_spec.rs
```diff
@@ -34,7 +34,7 @@ use sp_runtime::{
 	bounded_vec,
 	traits::{IdentifyAccount, Verify},
 };
-const SAFE_XCM_VERSION: u32 = xcm::v2::VERSION;
+const SAFE_XCM_VERSION: u32 = xcm::latest::VERSION;
```
Hounsette, line 37: wouldn't it be better to have the fixed version here to prevent future updates of xcm to be unwillingly brought to the codebase?

runtimes/eden/Cargo.toml
```diff
@@ -25,7 +25,7 @@ std = [
   "pallet-membership/std",
   "pallet-multisig/std",
   "pallet-offences/std",
-  "pallet-randomness-collective-flip/std",
+  "pallet-insecure-randomness-collective-flip/std",
```
ETeissonniere, line 28: Do we have an alternative to this pallet provided by Parity?

runtimes/eden/src/pallets_governance.rs
```diff
@@ -94,6 +94,7 @@ impl pallet_collective::Config<pallet_collective::Instance1> for Runtime {
 	type WeightInfo = ();
 	type MaxMembers = MaxMembers;
 	type DefaultVote = pallet_collective::PrimeDefaultVote;
+	type SetMembersOrigin = EnsureRoot<AccountId>;
```
ETeissonniere, line 97:
```suggestion
	type SetMembersOrigin = EnsureRootOrMoreThanHalfOfTechComm;
```

runtimes/eden/src/pallets_util.rs
```diff
@@ -173,9 +173,9 @@ impl pallet_contracts::Config for Runtime {
 	type DeletionQueueDepth = DeletionQueueDepth;
 	type DeletionWeightLimit = DeletionWeightLimit;
 	type Schedule = MySchedule;
-	type CallStack = [Frame<Self>; 31];
+	type CallStack = [Frame<Self>; 5];
 	type AddressGenerator = pallet_contracts::DefaultAddressGenerator;
-	type MaxCodeLen = ConstU32<{ 128 * 1024 }>;
+	type MaxCodeLen = ConstU32<{ 123 * 1024 }>;
```
ETeissonniere, line 178: Any reasons why this is getting lowered?

ETeissonniere (commented):
Some minor comments, looking good otherwise

node/src/chain_spec.rs
aliXsed: @Hounsette The latest version of XCM would only be considered safe if we have done a full upgrade and supports that version. Bringing the safe version from a particular version was the source of some tricky upgrade issues.

runtimes/eden/Cargo.toml
aliXsed: @ETeissonniere SRLabs have pointed us to some alternatives (although as a low priority issue) which we will be looking into as the next task. This is just a renaming for the same package at this stage and some system parachains that I have checked are still relying on it.

runtimes/eden/src/pallets_util.rs
aliXsed: I follow the contracts system parachain and the fact that they have gone for a lower value here could be due to some safety analysis. We'd better go for the lower and increase it in the future when we have evaluated all the cons and pros.

ETeissonniere: Yes of course - I was wondering as to _why_ they did the change, not necessarily _if_ we should do it

ETeissonniere: Looks like Parity's team figured it would be appropriate to document the limits and safe values in the not so well known `integrity_test` function. https://paritytech.github.io/substrate/master/src/pallet_contracts/lib.rs.html#323-386

node/src/chain_spec.rs
ETeissonniere: @aliXsed using the change below might allow us to avoid _some_ border effects in case we have an oversight.
```suggestion
const SAFE_XCM_VERSION: u32 = xcm::v3::VERSION;
```

ETeissonniere (approved):

node/src/chain_spec.rs
aliXsed: Just for reference, we discussed the cons and pros of going for the latest. Since we have had other places where the xcm definitions are fetched from the latest and we don't see any particular safety issue right now based on supporting the latest xcm that has officially brought to the scope of out code base, we decided to leave this here as it is.

PR #740, opened 2023-05-09
feat: add nodle uniques which is going to extend on substrate uniques

aliXsed:
We want to allow an adjustable intrinsic value for NFTs on top of [what is fixed as the base intrinsic value](https://github.com/NodleCode/chain/pull/740/files#diff-bc781f5a3dd945b1c46c426a837c2bd87d1f19a31ba63080b52f1d615575951dL110-L113) in the runtime configuration. 
In order to make sure, all the substrate uniques calls go through this new pallet, we are going to remove the direct Call access to the substrate uniques pallet. We then aim to cover all the main substrate uniques calls wrapped in our own Nodle uniques calls. This is so we make sure no intrinsic values go lost and the proper accounting will happen on all related functions such as mint, destroy or transfer. The intrinsic value could then be reserved in an owner's account and repatriated on transfers and unreserved on destroy (similar to what is done for the fixed deposit for an NFT where it reserves the deposit ([1](https://github.com/paritytech/substrate/blob/polkadot-v0.9.40/frame/uniques/src/functions.rs#L79), [2](https://github.com/paritytech/substrate/blob/polkadot-v0.9.40/frame/uniques/src/functions.rs#L172)), [unreserve it](https://github.com/paritytech/substrate/blob/polkadot-v0.9.40/frame/uniques/src/functions.rs#L131) and [repatriate it](https://github.com/paritytech/substrate/blob/polkadot-v0.9.40/frame/uniques/src/lib.rs#L872C17-L878) on different calls).

- [ ] Similar to how it's done for the [Create extrinsic](https://github.com/NodleCode/chain/pull/740/files#diff-1576dccf99d110eae22e43b44c85d93777af4108c57eb7053020525880da3aadR88-R96), wrap and present the following calls in pallet_nodle_uniques:
- - [ ] `destroy`
- - [ ] `mint`
- - [ ] `burn`
- - [ ]  `transfer`

- [ ] Add `create_with_extra_deposit` and `mint_with_extra_deposit` extrinsics
- [ ] Add tests for the accounting of the extra deposit on all possible extrinsics
- [ ] Add benchmarks for pallet_nodle_uniques (you can get inspiration or use the same benchmarks tests as those for pallet_uniques
- [ ] Add migration to make sure the storage of the pallet_uniques which was called Uniques is now prefixed with SubstrateUniques instead and available through this new name.

**NOTE**: This is an initial draft showcasing the overall approach.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6005,6 +6005,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "pallet-nodle-uniques"
+version = "2.1.0"
+dependencies = [
+ "frame-benchmarking",
+ "frame-support",
+ "frame-system",
+ "parity-scale-codec",
+ "scale-info",
+ "serde",
+ "sp-io",
+ "sp-runtime",
+ "sp-std",
+]
+
+[[package]]
 name = "pallet-nomination-pools"
 version = "1.0.0"
 source = "git+https://github.com/paritytech/substrate?branch=polkadot-v0.9.40#76fed9b9082daade5392663f25ed8968f8e8c11c"
@@ -8813,6 +8828,7 @@ dependencies = [
  "pallet-mandate",
  "pallet-membership",
  "pallet-multisig",
+ "pallet-nodle-uniques",
  "pallet-offences",
  "pallet-preimage",
  "pallet-reserve",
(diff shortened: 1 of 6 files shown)


aliXsed:
Closing this in favour of #742

PR #741, opened 2023-05-09
(fix)cargo doc workflow

Hounsette:


ETeissonniere (approved):

.github/workflows/doc.yml
```diff
@@ -3,20 +3,13 @@ name: Rust AutoDoc
...
 jobs:
   autodoc:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          toolchain: ${{ env.nightly }}
-          target: ${{ env.target }}
+      - run: rustup show
```
aliXsed, line 11: By this are we relying on the version of the rust that is preinstalled on the ubuntu-latest machine?

Hounsette: more reading [here](https://github.com/rust-lang/rust/issues/93333). 
It will be overridden by the version declared in the rust-toolchain file. 

We can replace all the actions-rust-lang/setup-rust-toolchain@v1 in the workflow files by rustup show

aliXsed: Hey @Hounsette. The link you have put here seems not relevant to our discussion or I couldn't find out how. Also I have noticed after the merge of this PR some builds have failed. I still don't get how `rustup show` can replace setting up the toolchain if it's not already setup.

Hounsette: It was the wrong link,  [here](https://github.com/actions-rs/toolchain/issues/126#issuecomment-782989659) is the explanation.

PR #746, opened 2023-05-18
Chore/20230518 release-build

simonsso:
This is now able to create a release automatically from a tag

Example release is generated here: https://github.com/NodleCode/chain/releases/tag/untagged-4b4bd7c5ae73fbfd223a

### DIFF
--- a/.github/workflows/srtool.yml
+++ b/.github/workflows/srtool.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    tags:
+      - "v*"
 
 jobs:
   srtool:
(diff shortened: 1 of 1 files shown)


.github/workflows/srtool.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    tags:
+      - "[0-9].*"
```
Hounsette, line 8: does it mean that this workflow will be triggered on a tag creation?

simonsso: Yes that is correct

simonsso: We have used tags of the 2.0.30 format so I continued that

Hounsette (approved):

PR #749, opened 2023-05-25
Use `runtime_version` attribute macro

bkchr:
This macro puts the `RuntimeVersion` into a custom section in the wasm file. This speedups reading the `RuntimeVersion`.


simonsso:
Thanks

PR #751, opened 2023-06-05
missions registry

ETeissonniere:
A simple first stab at what will become our smart missions registry. Necessary for the Polkadot Decoded live demo in Copenhagen.

#### TODOs
- [ ] add unit tests
- [ ] add benchmarks
- [ ] figure out the right way to include this in the repo without deploying to production yet


ETeissonniere:
We ended up not needing this.

PR #753, opened 2023-06-16
Pallet uniques: Mint with extra deposit

Hounsette:
- [x] Add create collection with extra deposit limit
- [x] Add update extra deposit limit for existing collections
- [x] Add mint with extra deposit
- [x] Unreserve the extra deposit on burning an item and transfer it to item owner
- [x] Unreserve all extra deposits on destroying a collection and transfer them to item owners
- [x] Repatriate the total extra deposit when the ownership of collection changes
- [x] Benchmark nodle-uniques-pallet
- [x] Relocate Uniques to SubstrateUniques pallet on runtime upgrade


pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,755 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! Handle the ability to notify other pallets that they should stop all
+
+use frame_support::traits::{Currency, ExistenceRequirement, ReservableCurrency};
+pub use pallet::*;
+use pallet_uniques::{DestroyWitness, WeightInfo as UniquesWeightInfo};
```
simonsso, line 25: This import is difficult to read. It is used as 
```
      #[pallet::weight(<T as pallet_uniques::Config<I>>::WeightInfo::buy_item())]

```

aliXsed: You need to bring the trait `WeightInfo` from the pallet_uniques into the scope without a name collision with this pallet's own WeightInfo. That's the solution we have.

launch.json
```diff
@@ -14,7 +14,7 @@
                 "port": 30555
             },
             {
-                "name": "charlie",
+                "name": "charle",
```
aliXsed, line 17: Why do we need this name change?

Hounsette: We don't... fixing it

ETeissonniere (commented):
Some changes here and there may be necessary. It would be nice if the benchmarking changes are moved out of this PR too

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,754 @@
...
+			// Unreserve and transfer extra reserved deposit to the item owners
+			for (item_owner, extra_deposit) in item_owners {
+				<T as pallet_uniques::Config<I>>::Currency::unreserve(&collection_owner, extra_deposit);
+				<T as pallet_uniques::Config<I>>::Currency::transfer(
+					&collection_owner,
+					&item_owner,
+					extra_deposit,
+					ExistenceRequirement::AllowDeath,
+				)?;
+			}
```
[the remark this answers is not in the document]
aliXsed: @ETeissonniere this is implementing what we discussed here https://nodlecash.slack.com/archives/C02H6C0D4H1/p1686878129340799

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,754 @@
...
+		/// Weight: `O(1)`
+		#[pallet::call_index(11)]
+		#[pallet::weight(<T as pallet_uniques::Config<I>>::WeightInfo::transfer_ownership())]
+		pub fn transfer_ownership(
+			origin: OriginFor<T>,
+			collection: T::CollectionId,
+			owner: AccountIdLookupOf<T>,
+		) -> DispatchResult {
+			pallet_uniques::Pallet::<T, I>::transfer_ownership(origin, collection, owner)
+		}
```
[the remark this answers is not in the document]
aliXsed: @ETeissonniere since we use the ownership details only at the time an item is being burnt or a collection is being destroyed, it would work for us without any changes. In other words we rely on the ownership from the substrate uniques. When it's transferred, it's transferred for us too.

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,754 @@
...
+		///
+		/// It will still return an error in the case that the collection is unknown of the signer
+		/// is not permitted to call it.
+		///
+		/// Weight: `O(items.len())`
+		#[pallet::call_index(6)]
+		#[pallet::weight(<T as pallet_uniques::Config<I>>::WeightInfo::redeposit(items.len() as u32))]
+		pub fn redeposit(origin: OriginFor<T>, collection: T::CollectionId, items: Vec<T::ItemId>) -> DispatchResult {
+			pallet_uniques::Pallet::<T, I>::redeposit(origin, collection, items)
+		}
```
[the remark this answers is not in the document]
aliXsed: @ETeissonniere This only deals with the original deposit per item and doesn't impact on any information that we keep for the extra deposit.

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,754 @@
...
+			let mut item_owners: Vec<(T::AccountId, BalanceOf<T, I>)> = Vec::new();
+
+			// Recover the item owners for each item in the ExtraDeposit Storage
+			for (item, extra_deposit) in ExtraDeposit::<T, I>::iter_prefix(collection) {
+				if let Some(item_owner) = pallet_uniques::Pallet::<T, I>::owner(collection, item) {
+					item_owners.push((item_owner, extra_deposit));
+				}
+			}
+
+			let ret = pallet_uniques::Pallet::<T, I>::destroy(origin, collection, witness)?;
```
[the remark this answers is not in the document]
aliXsed: @ETeissonniere this function is now refactored. Please check the updated code.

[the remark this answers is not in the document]
aliXsed: @Hounsette would better explain why it was written like this before. I believe the aim was that if the destroy is going to fail we don't make a change to the extra deposit. However as the extrinsic is operational by default, even if destroy fails the change we make to the extra deposit will be reverted too.

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,754 @@
...
+				if let Some(item_owner) = item_owner {
+					<T as pallet_uniques::Config<I>>::Currency::unreserve(&collection_owner, extra_deposit);
+					<T as pallet_uniques::Config<I>>::Currency::transfer(
+						&collection_owner,
+						&item_owner,
+						extra_deposit,
+						ExistenceRequirement::AllowDeath,
+					)?;
+				}
+			}
```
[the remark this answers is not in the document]
aliXsed: Same here the original intent was we don't proceed to make any change to the storage before the burn function from substrate uniques does its checks. However the early return gain here could be trivial. I refactored this function and covered a couple of more error cases.

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,743 @@
...
+		#[transactional]
+		pub fn mint_with_extra_deposit(
+			origin: OriginFor<T>,
+			collection: T::CollectionId,
+			item: T::ItemId,
+			owner: AccountIdLookupOf<T>,
+			deposit: BalanceOf<T, I>,
+		) -> DispatchResult {
+			pallet_uniques::Pallet::<T, I>::mint(origin, collection, item, owner).and_then(|_| {
+				let collection_owner =
```
simonsso, line 735: Should the reserve not be on owner, the beneficiary?

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,743 @@
...
+			pallet_uniques::Pallet::<T, I>::buy_item(origin, collection, item, bid_price)
+		}
+
+		/// Mint an item of a particular collection with extra deposit.
+		///
+		/// The origin must be Signed and the sender must be the Issuer of the `collection`.
+		///
+		/// - `collection`: The collection of the item to be minted.
+		/// - `item`: The item value of the item to be minted.
+		/// - `owner`: The initial owner of the minted item.
```
simonsso, line 719:
```suggestion
		/// - `beneficiary`: The initial owner of the minted item.
```

Rename `owner` to `beneficiary` in the code too  This wording is used in documentation upstream (but not the code)

pallets/uniques/src/lib.rs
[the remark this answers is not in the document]
ETeissonniere: You are very correct - this slipped out of my mind

pallets/uniques/src/lib.rs
aliXsed: If we want to go even further in making this operation secure is that the collection owner sets a maximum extra deposit that minter can use in total for minting the items of this collection.

aliXsed: Probably the total_extra_deposit per collection by itself can do the job. When it's zero which is the default, no permission to mint with extra deposit.

ETeissonniere: I like the idea of setting a cap and think this would be a light change for a higher level of robustness.

We definitely do not want to reserve tokens from the _item owner_ as this would open room for abuse and edge cases. It looks like it would make sense to keep the deposit in the collection owner account up until the NFT is burnt (which effectively leads to the item owner _claiming_ the deposit).

aliXsed: Okay, I'd be implementing the cap shortly. CC @Hounsette

aliXsed: Here is the commit that adds the discussed security feature: https://github.com/NodleCode/chain/pull/753/commits/9c6d8d888a6f111a671ecc093c0bb39a7901b1a6

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,849 @@
...
+type AccountIdLookupOf<T> = <<T as frame_system::Config>::Lookup as StaticLookup>::Source;
+pub type BalanceOf<T, I = ()> =
+	<<T as pallet_uniques::Config<I>>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
+
+#[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, Default, TypeInfo, MaxEncodedLen)]
+pub struct ExtraDepositDetails<T: Balance> {
+	/// The cap for the total extra deposit for the collection
+	///
+	/// `total` should never go higher than `limit`.
+	limit: T,
```
ETeissonniere, line 50: Any reason for doing it that way VS a maximum deposit on a _per item_ basis?

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,849 @@
...
+				<T as pallet_uniques::Config<I>>::Currency::unreserve(&collection_owner, extra_deposit);
+				<T as pallet_uniques::Config<I>>::Currency::transfer(
+					&collection_owner,
+					&item_owner,
+					extra_deposit,
+					ExistenceRequirement::AllowDeath,
+				)
+			} else {
+				Ok(())
+			}
```
ETeissonniere, line 287: Shouldn't this decrease the `total` as well?

runtimes/eden/src/migrations.rs
```diff
@@ -0,0 +1,57 @@
...
+#[cfg(feature = "try-runtime")]
+use codec::{Decode, Encode};
+#[cfg(feature = "try-runtime")]
+use frame_support::Blake2_128Concat;
+#[cfg(feature = "try-runtime")]
+use pallet_uniques::CollectionDetails;
+#[cfg(feature = "try-runtime")]
+use primitives::{AccountId, Balance};
+#[cfg(feature = "try-runtime")]
+use sp_std::prelude::*;
```
ETeissonniere, line 13: If you want to drop the multiple lines of feature flags, I believe you can do something like the below
```suggestion
#[cfg(feature = "try-runtime")]
use {
  codec::{Decode, Encode},
  frame_support::Blake2_128Concat,
  pallet_uniques::CollectionDetails,
  primitives::{AccountId, Balance},
  sp_std::prelude::*
};
```

scripts/init.sh
```diff
@@ -5,8 +5,7 @@ set -e
 echo "*** Initializing WASM build environment"
 
 if [ -z $CI_PROJECT_NAME ] ; then
-   rustup update nightly
    rustup update stable
 fi
 
-rustup target add wasm32-unknown-unknown --toolchain nightly
\ No newline at end of file
+rustup target add wasm32-unknown-unknown --toolchain stable
```
ETeissonniere: Shall we update the rust-toolchain file ?

pallets/uniques/src/lib.rs
aliXsed: @ETeissonniere the number of items are not limited. So it wouldn't give control to the collection owner over how much in total they are allowing the issuer to put on the minted items.

pallets/uniques/src/lib.rs
aliXsed: No it shouldn't. Because it again messes with the maximum that the collection owner allows to be spent on that collection. If someone burns a token they receive its intrinsic value.

pallets/uniques/src/lib.rs
ETeissonniere: Very correct

scripts/init.sh
aliXsed: @simonsso what do you suggest?

runtimes/eden/src/migrations.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[cfg(feature = "try-runtime")]
+const UNIQUES_CLASS_PREFIX: &[u8] = b"Class";
+
+const NEW_UNIQUES_PALLET_NAME: &[u8] = b"SubstrateUniques";
+const OLD_UNIQUES_PALLET_NAME: &[u8] = b"Uniques";
+
+pub struct MovePalletUniquesToSubstrateUniques;
+impl OnRuntimeUpgrade for MovePalletUniquesToSubstrateUniques {
+	fn on_runtime_upgrade() -> Weight {
```
simonsso, line 24: How does this trigger...

- [ ] check trigger
- [ ] run and upgrade on local testnet

pallets/uniques/Cargo.toml
```diff
@@ -0,0 +1,49 @@
+[package]
+name = "pallet-nodle-uniques"
+version = "2.1.0"
+authors = ["Eliott Teissonniere <git.eliott@teissonniere.org>"]
```
simonsso, line 4:
```suggestion
authors = ["Nodle <support@nodle.com>"]
```

pallets/uniques/src/benchmarking.rs
```diff
@@ -0,0 +1,181 @@
+// This file is part of Substrate.
```
simonsso, line 1: What file did we copy? Did not all the code change?

aliXsed: It should be removed.

pallets/uniques/src/benchmarking.rs
```diff
@@ -0,0 +1,181 @@
...
+const SEED: u32 = 0;
+
+fn get_config<T: Config<I>, I: 'static>() -> (T::CollectionId, T::AccountId, AccountIdLookupOf<T>) {
+	let collection_owner: T::AccountId = account("collection_owner", 0, SEED);
+	let collection_owner_lookup = T::Lookup::unlookup(collection_owner.clone());
+	let collection_id = <T as pallet_uniques::Config<I>>::Helper::collection(0);
+	(collection_id, collection_owner, collection_owner_lookup)
+}
+
+fn create_collection<T: Config<I>, I: 'static>(
```
simonsso, line 40: - [ ] There is only one create collection function not two versions

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,850 @@
...
+#[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, Default, TypeInfo, MaxEncodedLen)]
+pub struct ExtraDepositDetails<T: Balance> {
+	/// The cap for the total extra deposit for the collection
+	///
+	/// `total` should never go higher than `limit`.
+	limit: T,
+	/// The total extra deposit reserved so far
+	///
+	/// Ever mint with extra deposit should update this value.
+	total: T,
```
simonsso, line 54: Do we need to store both `total` and `limit`  one alternative could be `remaining`

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,850 @@
...
+	#[pallet::storage]
+	pub(crate) type CollectionExtraDepositDetails<T: Config<I>, I: 'static = ()> =
+		StorageMap<_, Blake2_128Concat, T::CollectionId, ExtraDepositDetails<BalanceOf<T, I>>, OptionQuery>;
+
+	#[pallet::error]
+	pub enum Error<T, I = ()> {
+		/// Issuing with extra deposit exceeds the collection owner's limit.
+		ExceedExtraDepositLimitForCollection,
+		/// The owner of collection is not known.
+		UnknownCollectionOwner,
```
simonsso, line 124: - [ ] Unknown is that the correct error here?

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,850 @@
...
+		/// - `collection`: The identifier of the new collection. This must not be currently in use.
+		/// - `admin`: The admin of this collection. The admin is the initial address of each
+		/// member of the collection's admin team.
+		///
+		/// Emits `Created` event when successful.
+		///
+		/// Weight: `O(1)`
+		#[pallet::call_index(0)]
+		#[pallet::weight(<T as pallet_uniques::Config<I>>::WeightInfo::create())]
+		pub fn create(
```
simonsso, line 149: create with no extra deposit

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,850 @@
...
+			witness.attributes,
+ 		))]
+		pub fn destroy(
+			origin: OriginFor<T>,
+			collection: T::CollectionId,
+			witness: DestroyWitness,
+		) -> DispatchResultWithPostInfo {
+			let collection_owner = pallet_uniques::Pallet::<T, I>::collection_owner(collection)
+				.ok_or(Error::<T, I>::UnknownCollectionOwner)?;
+			for (item, extra_deposit) in ItemExtraDeposits::<T, I>::drain_prefix(collection) {
```
simonsso, line 212: - [ ] Is there a testcase destroying one collection and keeping others?

pallets/uniques/src/lib.rs
```diff
@@ -0,0 +1,850 @@
...
+				let item_owner =
+					pallet_uniques::Pallet::<T, I>::owner(collection, item).ok_or(Error::<T, I>::UnknownItemOwner)?;
+				<T as pallet_uniques::Config<I>>::Currency::unreserve(&collection_owner, extra_deposit);
+				<T as pallet_uniques::Config<I>>::Currency::transfer(
```
simonsso, line 216: I think extra deposit should be owned by owner as reserved tokens, shouldn't it?

simonsso (commented):
Some comments and questions upto line 217
(I will check the checkboxen later)

scripts/init.sh
simonsso: Rust toolchain file is updated but it could not modify the apt-get installed version when running benchmarks so this is only a reflection we nolonger need nightly

simonsso: I left this in because I was not sure if it was still in use in some corner of CI or something.

pallets/uniques/src/lib.rs
aliXsed: When there is no collection, there is no reason for keeping a fund reserved. This is the unreserve condition. Otherwise we may lock tokens forever

pallets/uniques/src/lib.rs
aliXsed: It would be useful to be able to query the total and limit separately. Otherwise for knowing how much is the total reserved for a collection we have to check and add all the items extra deposits through potentially several queries.

pallets/uniques/src/lib.rs
aliXsed: https://github.com/NodleCode/chain/pull/753/commits/b4753fd265f743a4d6dbb9c3f1c0eac3c6ba8f52 implements it

pallets/uniques/src/benchmarking.rs
aliXsed: https://github.com/NodleCode/chain/pull/753/commits/798601553e8565df5de74c276725c8e417ca9605 implements it

ETeissonniere (approved):

PR #756, opened 2023-06-21
Use new name for artifact

simonsso:


simonsso:
This should trigger a try runtime build with the old config.

PR #757, opened 2023-06-22
Add GH secret for API key to try-runtime job

simonsso:


.github/workflows/docker.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         if: github.event_name != 'pull_request'
         uses: docker/login-action@v2 
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          username: ${{ vars.DOCKERHUB_USERNAME }}
```
simonsso, line 49: Moved username from secrets so it can be printed

simonsso:
Looks good to me

PR #764, opened 2023-07-05
Update benchmark script

simonsso:
Broke out the benchmark changes from PR #753


pallets/reserve/src/lib.rs
```diff
@@ -20,6 +20,7 @@
 
 //! A module that is called by the `collective` and is in charge of holding
 //! the company funds.
+#[cfg(feature = "runtime-benchmarks")]
```
ETeissonniere, line 23: I do not think this necessary as it is already at the top of `benchmarking.rs`?

pallets/reserve/src/benchmarking.rs
```diff
@@ -37,15 +37,19 @@ const SEED: u32 = 0;
...
+		let starting_amount = smallest_usable_amount.saturating_add(smallest_usable_amount);
+		let tip_amount = smallest_usable_amount;
+		let _ = T::Currency::make_free_balance_be(&tipper,starting_amount);
+	}: _(RawOrigin::Signed(tipper), tip_amount)
 
 	spend {
 		let dest = account("dest", 0, SEED);
-		let value = 10u32.into();
+		let smallest_usable_amount = T::Currency::minimum_balance();
+		let starting_amount = smallest_usable_amount.saturating_add(smallest_usable_amount);
```
ETeissonniere, line 49: It appear the original code wasn't created a balance superior to amount spent. As such, I do not think we need this line of code

simonsso: Existing Deposits is enforced in tests now, I think the old code was broken.

I think it should spend some tokens here and update one storage item to get a heavier weight.

pallets/reserve/src/benchmarking.rs
```diff
@@ -37,15 +37,19 @@ const SEED: u32 = 0;
...
+		let value = smallest_usable_amount;
```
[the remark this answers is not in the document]
simonsso: Removed

pallets/reserve/src/lib.rs
simonsso: I will double check

aliXsed (approved):

PR #769, opened 2023-08-25
Keep Substrate Uniques in its original place and temporarily avoid introducing nodle_uniques

aliXsed:
- [x] Use `codeSubtitues` in eden chain spec to overrride the problematic runtime


ETeissonniere (changes requested):
`eden.json` must be updated to include the correct block as substitute as well as this is the file we use to generate the chain spec when compiling the node. `*.spec.json` files are the source files that were used to generate the raw specs but they are no longer really used I believe. In fact, we may want to delete them to avoid confusion in a future PR.

ETeissonniere (approved):

PR #768, opened 2023-08-23
hotfix: take out migration code

ETeissonniere:
- rm intrinsic nfts feature for now
- restore launch.json

PR #770, opened 2023-08-29
deprecate try-runtime feature in favor of new CLI from Parity

ETeissonniere:
- [x] confirm `try-runtime` actually works as intended
- [x] remove `try-runtime` command
- [x] check if we need to maintain or remove the `try-runtime` feature flags in the runtime

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,7 +131,7 @@ jobs:
           path: snapshots
           key: ${{steps.get_version.outputs.eden_rev}}
       - name: Run try-runtime
-        uses: NodleCode/action-try-runtime@78f23c43b7fa7c89924b148de669c6206bcaaf57
+        uses: NodleCode/action-try-runtime@518ad6cd5b2c81999e440f8d61bfccfafbd87add
         with:
           url: ${{ env.try-runtime-uri}}${{ secrets.DWELLIR_API_KEY}}
           snap: snapshots/eden-snapshot-full
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,6 +130,11 @@ jobs:
         with:
           path: snapshots
           key: ${{steps.get_version.outputs.eden_rev}}
+      - name: Ensure cache directory is created
+        if: steps.cachedir.outputs.cache-hit != 'true'
+        run: |
+          install -d snapshots
+          date > snapshots/created_at
       - name: Run try-runtime
         uses: NodleCode/action-try-runtime@62a47e64e789d1c524d6c3e3c6ef36880a76901f
         with:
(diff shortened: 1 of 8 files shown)


aliXsed:
@ETeissonniere I wonder how this way of calling try-runtime can actually run the pre_upgrade and post_upgrade tests that we only build for the try-runtime feature: `#[cfg(feature = "try-runtime")]`? For a recent example of that, have a look at [here](https://github.com/NodleCode/chain/pull/768/files#diff-fbec76f42d30d207ff57eed6e555a0991778585832f45399cba9806e36cb8647L30-L38).

ETeissonniere:
> @ETeissonniere I wonder how this way of calling try-runtime can actually run the pre_upgrade and post_upgrade tests that we only build for the try-runtime feature: #[cfg(feature = "try-runtime")]? For a recent example of that, have a look at [here](https://github.com/NodleCode/chain/pull/768/files#diff-fbec76f42d30d207ff57eed6e555a0991778585832f45399cba9806e36cb8647L30-L38).

We still test with a try runtime enabled runtime and as such the right code can still be run. See below:
```
      # This build the runtime with try-runtime turned on
      - name: Cargo build
         run: cargo build --release --features=try-runtime -p runtime-eden
       # [... abbreviated ...]
       # Note how `--runtime` point towards the runtime we just built
       - name: Try runtime reuse snap
         run: |
             cat snapshots/created_at
             try-runtime --runtime target/release/wbuild/runtime-eden/runtime_eden.wasm on-runtime-upgrade --checks=all snap --path snapshots/eden-snapshot-full
```

.github/workflows/ci.yml
```diff
@@ -122,22 +122,23 @@ jobs:
...
-            date > snapshots/created_at
-            ./target/release/nodle-parachain try-runtime --runtime existing -lruntime=debug --chain ${{ env.try-runtime-chain }} create-snapshot "snapshots/eden-snapshot-full" -u ${{ env.try-runtime-uri}}${{ secrets.DWELLIR_API_KEY}}
-      - name: Try runtime reuse snap
-        run: |
-            cat snapshots/created_at
-            ./target/release/nodle-parachain try-runtime --runtime target/release/wbuild/runtime-eden/runtime_eden.wasm -lruntime=debug --chain ${{ env.try-runtime-chain }} on-runtime-upgrade --checks=all snap -s snapshots/eden-snapshot-full
+          install -d snapshots
+          date > snapshots/created_at
+      - name: Run try-runtime
+        uses: NodleCode/action-try-runtime@62a47e64e789d1c524d6c3e3c6ef36880a76901f
```
aliXsed, line 139: @ETeissonniere It's so nice that you take the action out to its own repository.  Can we we have a release of that repo and use it here instead of the commit?

aliXsed (approved):
Nice job, I have left only one minor comment, but it's now in a good shape.

.github/workflows/ci.yml
ETeissonniere: We will once it actually works

PR #771, opened 2023-08-29
Overhaul Github Actions to be faster

ETeissonniere:
**This is a WIP.**


ETeissonniere:
Having everything runs via docker as I originally intended seems overkill and suboptimal. Will investigate alternatives such as sccache.

PR #780, opened 2023-09-08
Remove nodle uniqes and migration

simonsso:


aliXsed:
It's enough that nodle uniques is not used in constructing our latest runtime. We should keep the pallet in the code.

PR #782, opened 2023-09-12
Remove 2.3.0 reverse migration and bump to 2.3.1

simonsso:
This is the 2.3.0 code intended for release on mainnet


simonsso:
Migration executed on testnet it can be removed now

PR #787, opened 2023-09-14
Update weights for sponsorship pallet

simonsso:


simonsso (commented):
LGTM only minor changes from yesterday without the fix Will merge after CI is ok

PR #792, opened 2023-10-17
forward errors when dispatch fail on reserve.apply_as

ETeissonniere:
This leads to batch_all also forwarding errors if `reserve.apply_as` fails.

See [this proposal](https://nodle.subscan.io/tech/313?tab=votes) for an example of a proposal which partially fails but still gets partially applied which is not a desired border effect.


aliXsed (approved):
LGTM

PR #796, opened 2023-10-26
get Nodle parachain ready for state trie migration

ETeissonniere:
It appears we never fully completed the State Trie migration from prior versions of Substrate. ~~Not doing so seems to be preventing correct usage of the Warp sync mode which could save us a lot of money on node storage and syncing time.~~ Additionally it appears the new State Trie format may help reducing the PoV size for Parachain blocks which is what posed us issues last month.

This PR adds the right RPC, pallet, and version changes for performing the state migration manually, potentially via [Parity's script](https://github.com/paritytech/polkadot-scripts/blob/master/src/services/state_trie_migration.ts). More information in [this great HackMD](https://hackmd.io/@kizi/HyoSO3lf9). The changes from this PR can be safely reverted post migration.

The entire migration flow which includes setting transaction limits and running the **full** migration via the prior mentioned script or a fork should be fully tested end to end on the testnet prior to main net usage. Nonetheless, it appears this is a relatively safe migration since Substrate will operate in "hybrid" mode (not warp compatible) by default until we complete the migration via the script. If the script, is not run, or simply does not work, the migration will be done opportunistically, meaning that Substrate will migrate storage values as they are accessed on the fly. (In fact the script / pallet combination simply iterates through every storage values by reading them as per my understanding - correct me if I am wrong).

### Remaining TODOs
- [ ] add benchmarks
- [x] find the right safe `MaxKeysLen` value

### Post deployment sanity checks
- ensure `utils_trieMigrationStatus` still return a `MaxKeyLen` of no higher than 120
- ensure we can run the signed migration via the bot for all of the chain without errors

In case of any failure we can simply make a new runtime upgrade to increase `MaxKeyLen`, or resume the migration after a script fix which wouldn't require a runtime upgrade.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5432,6 +5432,7 @@ dependencies = [
  "substrate-build-script-utils",
  "substrate-frame-rpc-system",
  "substrate-prometheus-endpoint",
+ "substrate-state-trie-migration-rpc",
  "xcm",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9264,6 +9264,7 @@ dependencies = [
  "pallet-scheduler",
  "pallet-session",
  "pallet-sponsorship",
+ "pallet-state-trie-migration",
  "pallet-timestamp",
  "pallet-transaction-payment",
  "pallet-transaction-payment-rpc-runtime-api",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5389,6 +5389,7 @@ dependencies = [
  "hex-literal 0.3.4",
  "jsonrpsee",
  "log",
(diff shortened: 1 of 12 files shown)


ETeissonniere:
After further analysis it looks like Polkadot and Statement did not yet migrate (unlike Kusama and Westend). As such this does not seem to be an immediate priority and invalidate my hypothesis that it would be any useful for our warp sync + pruning efforts.

Gonna revert this to draft state for archival purposes. This is not an immediate priority until both Polkadot and Statement do this same migration without issues.

PizzaShift (approved):
hello

PR #798, opened 2023-11-14
feat(sponsorship): add more details to sponsorship events

aliXsed:
The extra detail in the events is not going to impact the cost of the corresponding extrinsics sensibly. However they help with the performance of the indexers such as Subquery which follow the events and are better off just doing that and not needing to query a value from a node. 

- [x] @Douglasacost to review the extra details for the events and check if they are enough
- [x] Benchmark the pallet again so we can confirm the extrinsic costs would be minimally different.


pallets/sponsorship/src/lib.rs
```diff
@@ -157,19 +157,42 @@ pub mod pallet {
...
 		/// Event emitted when fee_quota or reserve_quota or both are updated for the given list
-		UsersLimitsUpdated { pot: T::PotId, users: Vec<T::AccountId> },
+		UsersLimitsUpdated {
+			pot: T::PotId,
+			users: Vec<T::AccountId>,
+			fee_quota: BalanceOf<T>,
+			reserve_quota: BalanceOf<T>,
+		},
 		/// Event emitted when a sponsor_me call has been successful indicating the reserved amount
 		Sponsored { paid: BalanceOf<T>, repaid: BalanceOf<T> },
```
Douglasacost, line 197: What `repaid` means?

Douglasacost (approved):
LGTM!

pallets/sponsorship/src/lib.rs
aliXsed: @Douglasacost `paid` is how much was borrowed from the sponsor at the beginning of this call and `repaid` means how much debt was payed off from the user's account at the end of this call.

PR #801, opened 2023-11-24
feat(sponsorship): fight spam

aliXsed:
- [x] Ask for deposit for pot
 - [x] Limit num of users per pot
 - [x] Bound extrinsic computations
 - [x] Add migration
 - [x] Convert migration to multi-block partial migrations
 - [x] Cover the added logic by unit tests
 - [x] Benchmark migration
 - [x] Benchmark pallet
 - [x] Rebase to master
 
Closes [CHA-401](https://linear.app/nodle/issue/CHA-401) and [CHA-402](https://linear.app/nodle/issue/CHA-402)


ETeissonniere (changes requested):
@aliXsed if we adopt this strategy we also need to handle pots created before then, like our app's. This means some form of migration to make sure the `unreserve` does not have any unintended effects or errors.

runtimes/eden/src/pallets_util.rs
```diff
@@ -163,12 +163,18 @@ impl Default for SponsorshipType {
 	}
 }
 
+parameter_types! {
+	pub const PotDeposit: Balance = 1000 * constants::NODL;
+	pub const RegisterLimit: u32 = 200;
```
ETeissonniere, line 168: This needs to be substantially more. We are targeting millions of users for our upcoming app.

runtimes/eden/src/pallets_util.rs
```diff
@@ -163,12 +163,18 @@ impl Default for SponsorshipType {
 	}
 }
 
+parameter_types! {
+	pub const PotDeposit: Balance = 1000 * constants::NODL;
```
ETeissonniere, line 167: Probably a little cheap considering existing NODL price.

runtimes/eden/src/pallets_util.rs
aliXsed: How about a per user deposit instead. So we don't set any limitations on the number?

runtimes/eden/src/pallets_util.rs
aliXsed: It's just an example, can you tell me what would be a better price?

aliXsed:
> @aliXsed if we adopt this strategy we also need to handle pots created before then, like our app's. This means some form of migration to make sure the `unreserve` does not have any unintended effects or errors.

@ETeissonniere The unreserve is a no failure function. So we are safe on that side. The potential side effect is an unreserve for a fund that should be still reserved. But we know that the unreserve is not going to exceed the pot deposit. So not very consequential in practice. Also at this stage we only have 3 pots mainly created by ourselves I believe. IMO at this stage a complicated migration is not warranted.

runtimes/eden/src/pallets_util.rs
aliXsed: Based on our conversation today, let's keep this low so we do not limit any possible cool usage. We can  reevaluate later

runtimes/eden/src/version.rs
simonsso, line 1: This file should not be changed in this PR. (yes i know it is broken)

pallets/sponsorship/src/migration.rs
```diff
@@ -0,0 +1,216 @@
...
+		let mut fee_quota = LimitedBalance::with_limit(pre.fee_quota.limit());
+		fee_quota.saturating_add(pre.fee_quota.balance());
+		let mut reserve_quota = LimitedBalance::with_limit(pre.reserve_quota.limit());
+		reserve_quota.saturating_add(pre.reserve_quota.balance());
+		crate::PotDetailsOf::<T> {
+			sponsor: pre.sponsor,
+			sponsorship_type: pre.sponsorship_type,
+			fee_quota,
+			reserve_quota,
+			deposit: Default::default(),
```
ETeissonniere, line 66:
```suggestion
			deposit: Zero::zero(),
```

Probably more expressive than `Default`.

pallets/sponsorship/src/migration.rs
```diff
@@ -0,0 +1,216 @@
...
+	let translate_user_details = |pre: v0::UserDetailsOf<T>| -> crate::UserDetailsOf<T> {
+		let mut fee_quota = LimitedBalance::with_limit(pre.fee_quota.limit());
+		fee_quota.saturating_add(pre.fee_quota.balance());
+		let mut reserve_quota = LimitedBalance::with_limit(pre.reserve_quota.limit());
+		reserve_quota.saturating_add(pre.reserve_quota.balance());
+		crate::UserDetailsOf::<T> {
+			proxy: pre.proxy,
+			fee_quota,
+			reserve_quota,
+			deposit: Default::default(),
```
ETeissonniere, line 78:
```suggestion
			deposit: Zero::zero(),
```

Probably more expressive than `Default`.

pallets/sponsorship/src/lib.rs
```diff
@@ -349,7 +375,14 @@ pub mod pallet {
 					user_details.reserve_quota.balance(),
 				)?;
 				pot_details.reserve_quota.saturating_sub(repaid);
-				Self::remove_user(&pot, &user);
+				UserRegistrationCount::<T>::mutate(&user, |count| {
+					count.saturating_dec();
+					if count.is_zero() {
```
simonsso, line 380: if count was zero, and this was called in an unexpected way this will be called twice.

aliXsed: It doesn't happen as count must be at least 1 at this point. I will however make a slight change for that unlikely case.

pallets/sponsorship/src/lib.rs
```diff
@@ -506,6 +561,30 @@ pub mod pallet {
...
 		}
 	}
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
+			// NOTE: the following consts should be adjusted according to a safe weight limit which
+			// should be way below the total block weight limit.
+			const MAX_POTS_TO_MIGRATE: usize = 200;
+			const MAX_USERS_TO_MIGRATE: usize = 100;
+			migration::migrate_partially::<T>(MAX_POTS_TO_MIGRATE, MAX_USERS_TO_MIGRATE)
```
ETeissonniere, line 571: How do we verify these constants?

pallets/sponsorship/src/migration.rs
```diff
@@ -0,0 +1,342 @@
...
+						deposit: Default::default(),
+					},
+				)
+			})
+			.collect::<Vec<_>>();
+
+		let num_of_pots = pots.len();
+
+		pots.into_iter()
+			.for_each(|(pot, details)| V1Pot::<T>::insert(pot, details));
```
ETeissonniere, line 99: Can't this be done with only one iterator?

pallets/sponsorship/src/migration.rs
```diff
@@ -0,0 +1,342 @@
...
+			.collect::<Vec<_>>();
+
+		let users_len = users.len();
+
+		users.into_iter().for_each(|(pot, user, details)| {
+			UserRegistrationCount::<T>::mutate(&user, |count| {
+				count.saturating_inc();
+			});
+			V1User::<T>::insert(pot, user, details);
+		});
```
ETeissonniere, line 138: I wonder if this can be done in one iter

pallets/sponsorship/src/migration.rs
```diff
@@ -0,0 +1,342 @@
...
+			(None, weight)
+		}
+	}
+
+	pub fn migrate_partially<T: Config>(max_pots: usize, max_users: usize) -> Weight {
+		let mut weight: Weight = T::DbWeight::get().reads(2);
+
+		let pot_migration_in_progress = if let Some(starting_key) = PotMigrationCursor::<T>::get() {
+			let (end_cursor, migration_weight) = migrate_pot_partially::<T>(max_pots, starting_key);
+			weight += migration_weight;
```
ETeissonniere, line 155: Add one write for the storage changes below
```suggestion
			weight += migration_weight +  T::DbWeight::get().writes(1);
```

pallets/sponsorship/src/migration.rs
```diff
@@ -0,0 +1,342 @@
...
+					false
+				}
+			}
+		} else {
+			false
+		};
+
+		let user_migration_in_progress = if let Some(starting_key) = UserMigrationCursor::<T>::get() {
+			let (end_cursor, migration_weight) = migrate_user_partially::<T>(max_users, starting_key);
+			weight += migration_weight;
```
ETeissonniere, line 172: Add one write for the storage changes below
```suggestion
			weight += migration_weight +  T::DbWeight::get().writes(1);
```

pallets/sponsorship/src/migration.rs
aliXsed: The thing that might not be obvious here is that v0::Pot and V1Pot are both referencing the same storage. We shouldn't make any changes to one of them while iterating over another one. So I need to read from Pot, do my operation and make it sure it's applied (by collecting it) so I can then safely start to iterate and change the other one. Any suggestion is welcomed.

pallets/sponsorship/src/migration.rs
aliXsed: Same as my response to other comment

runtimes/eden/src/version.rs
Brycewetzel: Yes.

ETeissonniere (dismissed):
🔥

PR #806, opened 2023-12-15
Update cargo lock and rolback rev

simonsso:


runtimes/eden/src/version.rs
```diff
@@ -40,7 +40,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// Version of the runtime specification. A full-node will not attempt to use its native
 	// runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
 	// `spec_version` and `authoring_version` are the same between Wasm and native.
-	spec_version: 27,
+	spec_version: 26,
```
aliXsed, line 43: Please revert this change.

PR #810, opened 2023-12-19
(Do not merge) Remove migrations

simonsso:


runtimes/eden/src/version.rs
```diff
@@ -48,7 +48,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// do the same thing.
 	// Non-consensus-breaking optimizations are about the only changes that could be made which
 	// would result in only the `impl_version` changing.
-	impl_version: 0,
+	impl_version: 1,
```
aliXsed, line 51: The advice for substituting code was to keep all the versions the same (disclaimer: I haven't investigated the advice independently). So we need to make sure this is ok.

runtimes/eden/src/pallets_util.rs
```diff
@@ -219,9 +219,9 @@ impl pallet_contracts::Config for Runtime {
 	type UnsafeUnstableInterface = ConstBool<false>;
 	type MaxDebugBufferLen = ConstU32<{ 2 * 1024 * 1024 }>;
 	type Migrations = (
-		pallet_contracts::migration::v10::Migration<Runtime>,
-		pallet_contracts::migration::v11::Migration<Runtime>,
-		pallet_contracts::migration::v12::Migration<Runtime>,
+		pallet_contracts::NoopMigration<10>,
+		pallet_contracts::NoopMigration<11>,
+		pallet_contracts::NoopMigration<12>,
```
aliXsed, line 224: It's likely that the following change does the same job (other than debug printing the step):
```suggestion
```

simonsso: The `NoopMigratoin` calls  will set state to 12 when done. without it there will be no change.

runtimes/eden/src/version.rs
simonsso, line 51:
```suggestion
	impl_version: 0,
```

simonsso: Lets just roll back this as it is an increased risk for no major benefit

simonsso:
Released from feature branch https://github.com/NodleCode/chain/tree/recover_testnet_2_4_1_recover as https://github.com/NodleCode/chain/releases/tag/2.4.1-rc1

PR #811, opened 2023-12-19
(Do Not Merge)Iso/remove contracts storage

simonsso:


simonsso:
Strategy chosen is to do this by hand

PR #812, opened 2023-12-20
Add a lint for json files

simonsso:


.github/workflows/ci.yml
```diff
@@ -16,7 +16,15 @@ env:
   RUSTFLAGS: ""
 
 jobs:
-  lints:
+  json-lints:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check for json syntax errors 1
+        run: "for i in `git ls-tree --full-tree -r --name-only HEAD |grep json`; do echo Checking $i:; json_pp <$i >/dev/null || echo ERROR: $i failed; done"
```
simonsso, line 24: I know this is a hack, but until we figure out how to spin these off in a matrix build this will be good enough and the execution time will never justify even one hour spend on optimizing this.

simonsso:
@ETeissonniere an alternative solution to fixing the syntax violations in devcontainer.json is to add a preprocessor step to the check and remove all C++ syle line comments

.github/workflows/ci.yml
aliXsed: My suggestion is we shouldn't lint our json assets directly. Every json asset should be tested through their own consumer. For example eden.json is consumed by nodle-parachain. After building nodle-parachain, we need to test if nodle-parachain will be happy with eden.json and paradis.json as its chain input. Same for the dev container. If we are creating a dev container and its own process is happy with devcontainer.json format, that's what we need.

PR #822, opened 2024-02-01
chore: allow us to run some CI jobs manually when we need

aliXsed:
We will need to re-run try-runtime tests later for a specific tag (release) close to the deployment time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,15 @@ on:
     branches:
       - master
   pull_request:
-
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to run the workflow on'
+        required: true
+      run_specific_job:
+        description: 'Specify a job to run (test-runtime, try-runtime-execute, etc.)'
+        required: true
+  
 env:
   toolchain: stable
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -5,7 +5,15 @@ on:
...
   pull_request:
-
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to run the workflow on'
+        required: true
+      run_specific_job:
+        description: 'Specify a job to run (test-runtime, try-runtime-execute, etc.)'
+        required: true
```
ETeissonniere, line 15: I do not think this is necessary

aliXsed: You mean the whole workflow_dispatch?

simonsso: I would just have forked the tag and created a dummy PR, but if this works it is OK for me.

simonsso (approved):

aliXsed:
@simonsso there was one way to test if it works and that was through merging it.

aliXsed:
Here is how this PR has enabled the feature:
Screenshot 2024-02-09 at 5 53 30 PM

PR #833, opened 2024-03-11
Compound extinsics in sponsorship

simonsso:


runtimes/eden/src/pallets_util.rs
```diff
@@ -153,12 +153,21 @@ impl pallet_uniques::Config for Runtime {
 pub enum SponsorshipType {
 	AnySafe,
 	Uniques,
+	// Nft,
```
simonsso, line 156: I don't see why we need a AnySafe

runtimes/eden/src/pallets_util.rs
```diff
@@ -153,12 +153,21 @@ impl pallet_uniques::Config for Runtime {
...
 }
 impl InstanceFilter<RuntimeCall> for SponsorshipType {
 	fn filter(&self, c: &RuntimeCall) -> bool {
 		match self {
+			// SponsorshipType::Nft => matches!(c, RuntimeCall::Utility {  })
```
simonsso, line 161: we should have tests for this function

runtimes/eden/src/pallets_util.rs
aliXsed: We need to provide the option to allow any function that may find a use case to be sponsored. We'd just need to exempt unsafe (in terms of being able to violate the sponsorship contract), hence the name AnySafe.

pallets/sponsorship/src/benchmarking.rs
```diff
@@ -225,6 +225,16 @@ mod benchmarks {
 		}
 	}
 
+	#[benchmark]
+	fn fn_validate() {
```
aliXsed, line 229: Trivial: no need to start a fn with `fn_`

pallets/sponsorship/src/lib.rs
```diff
@@ -278,7 +279,7 @@ pub mod pallet {
 			reserve_quota: BalanceOf<T>,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
-			ensure!(!Self::migration_in_progress(), Error::<T>::MigrationInProgress);
+			Self::ensure_migration_completed()?;
```
aliXsed, line 282: We don't need to migration logic anymore, do we? Suggest you leave taking out the migration logic to another PR.

pallets/sponsorship/src/lib.rs
```diff
@@ -17,6 +17,7 @@
  */
 
 #![cfg_attr(not(feature = "std"), no_std)]
+#![deny(unused_must_use)]
```
aliXsed, line 20: I suppose this is temporary for your own test. Otherwise, the warning is the preferred behaviour in the case of unused return value.

pallets/sponsorship/src/lib.rs
```diff
@@ -787,7 +792,8 @@ where
 		info: &DispatchInfoOf<Self::Call>,
 		len: usize,
 	) -> TransactionValidity {
-		Self::validate_sponsor_call(who, call, info, len).map(|_| Ok(ValidTransaction::default()))?
+		Self::validate_sponsor_call(who, call, info, len)?;
+		Ok(ValidTransaction::default())
```
aliXsed, line 796: What's the motivation behind this change?

pallets/sponsorship/src/mock.rs
```diff
@@ -141,6 +149,7 @@ pub enum SponsorshipType {
 	Balances,
 	Uniques,
 	UniquesMint,
+	Compoud,
```
aliXsed, line 152: I had imagined supporting batch operations differently. In my view batching (or as you named compound) is not a sponsorship type. `batch_sponsor` will be a sponsorship function with the following signature: `pub fn batch_sponsor(origin: OriginFor<T>, calls: Vec<<T as Config>::RuntimeCall>) -> DispatchResultWithPostInfo` at the same level with `sponsor_for`. This added function will allow a list of calls to be executed in a similar fashion as it's in the [utility pallet](https://github.com/paritytech/polkadot-sdk/blob/polkadot-v1.6.0/substrate/frame/utility/src/lib.rs#L201-L241). What we want to achieve is "if the selected sponsorship type (say Uniques) allows certain calls then those calls can also be submitted as a batch.

pallets/sponsorship/src/lib.rs
simonsso: ehhh

simonsso:
#836

PR #834, opened 2024-03-15
Silly test to verify sanity of DOT constants

simonsso:


simonsso:
Trivial changes will merge when tests run

runtimes/eden/src/constants.rs
```diff
@@ -143,4 +143,9 @@ mod tests {
 		assert_eq!(deposit(0, 1000), 60000000000 as Balance);
 		assert_eq!(deposit(0xFFFF_FFFF, 0xFFFF_FFFF), 901943131950000000 as Balance);
 	}
+
+	#[test]
+	fn polkadot_constants() {
+		assert_eq!(100 * POLKADOT_CENT, POLKADOT_EXISTENTIAL_DEPOSIT);
```
aliXsed, line 149: A better improvement here is that instead of redefining Polkadot constants here, we use them from the polkdot-sdk directly.

aliXsed: https://linear.app/nodle/issue/CHA-453

PR #835, opened 2024-03-18
Improve stability for try-runtime flow

simonsso:
- Try runtime will take about one minute to check every time now. Before there could be a penalty of 10-15 minutes if the snap was not in the cache and if the test failed the cache would not have been updated so the next check would also be slow.
- Executor size for try-runtime is reduced from 8 core to standard. 
- doc generation is tested on every PR, and ignores warning. (Autodoc is not the place to notify us about warnings)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
           out-file-path: "snaps"
           extract: false
       - name: Run try-runtime
-        uses: NodleCode/action-try-runtime/check@beta
+        uses: NodleCode/action-try-runtime/check@v0.6.1
         with:
           snap: snaps/eden-snapshot-full
           runtime: target/release/wbuild/runtime-eden/runtime_eden.wasm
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -143,27 +143,20 @@ jobs:
...
+          out-file-path: "snaps"
+          extract: false
+      - name: Check files
         run: |
-          install -d snapshots
-          date > snapshots/created_at
+           md5sum snaps/eden-snapshot-full
       - name: Run try-runtime
-        uses: NodleCode/action-try-runtime@v0.5.1
+        uses: NodleCode/action-try-runtime/check@beta
```
simonsso, line 157:
```suggestion
        uses: NodleCode/action-try-runtime/check@v0.6.1
```

When merged and tagged.

simonsso:
image
This test case is removed and will never pass

trsmarc (approved):
Overall look good, I see linter warning comment was added by GH workflow. If there is noting to worry please try to put them in linter ignore list.

simonsso:
> Overall look good, I see linter warning comment was added by GH workflow. If there is noting to worry please try to put them in linter ignore list.

Those are on my todo list to fix soon.

PR #836, opened 2024-03-18
Sponsorship vector implementation

simonsso:
- [x] tests for positive case where all pass
- [x] test where some jobs pass but on fails and all is rolled back
- [x] benchmark

### DIFF
--- a/pallets/sponsorship/src/lib.rs
+++ b/pallets/sponsorship/src/lib.rs
@@ -432,16 +432,16 @@ pub mod pallet {
 		/// are needed during pre and post dispatching this call.
 		#[pallet::call_index(4)]
 		#[pallet::weight({
-			let dispatch_infos = calls.iter().map(|call| call.get_dispatch_info()).collect::<Vec<_>>();
-			let dispatch_weight = dispatch_infos.iter().map(|di| di.weight)
-				.fold(Weight::zero(), |total: Weight, weight: Weight| total.saturating_add(weight));
-			let dispatch_class = if dispatch_infos.iter().all(|di| di.class == DispatchClass::Operational) {
-				DispatchClass::Operational
-			} else {
-				DispatchClass::Normal
-			};
-			(dispatch_weight + < T as Config >::WeightInfo::pre_sponsor() + < T as Config >::WeightInfo::post_sponsor() + T::DbWeight::get().reads_writes(2, 2), dispatch_class, Pays::No)
-		})]
+		let dispatch_infos = calls.iter().map(|call| call.get_dispatch_info()).collect::<Vec<_>>();
+		let dispatch_weight = dispatch_infos.iter().map(|di| di.weight)
+			.fold(Weight::zero(), |total: Weight, weight: Weight| total.saturating_add(weight));
+		let dispatch_class = if dispatch_infos.iter().all(|di| di.class == DispatchClass::Operational) {
+			DispatchClass::Operational
+		} else {
+			DispatchClass::Normal
+		};
+		(dispatch_weight + < T as Config >::WeightInfo::pre_sponsor() + < T as Config >::WeightInfo::post_sponsor() + T::DbWeight::get().reads_writes(2, 2), dispatch_class, Pays::No)
+	})]
 		pub fn sponsor_for(
 			origin: OriginFor<T>,
 			pot: T::PotId,
--- a/pallets/sponsorship/src/lib.rs
+++ b/pallets/sponsorship/src/lib.rs
@@ -404,8 +404,10 @@ pub mod pallet {
 
 		/// Sponsor me for the given call from the specified pot.
 		/// The caller must be registered for the pot.
-		/// The call must be consistent with the pot's sponsorship type.
+		/// The calls must be consistent with the pot's sponsorship type.
 		///
+		/// Calls: one or more extrinsics which will be executed paid by the sponsorship. If any call fails all will be rolled back
+
 		/// Returns Error if the pot doesn't exist or the user is not registered for the pot or if
 		/// their call is not matching the sponsorship type in which case the error would be
 		/// `frame_system::Error::CallFiltered`. Also returns error if the call itself should fail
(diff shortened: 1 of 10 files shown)


simonsso:
Batch mode, I think we only need to support **batch-all** mode from utility pallet. 

- **batch_all**: Stops on error and reverts everything (all-or-nothing).
- **force_batch**: Does not stop on error and only reverts the failed calls (keep-going).
- **batch**: Stops on error and only reverts the first failed call (fail-fast).
[ref](https://substrate.stackexchange.com/a/10403/116)

pallets/sponsorship/src/lib.rs
```diff
@@ -430,19 +432,29 @@ pub mod pallet {
 		/// are needed during pre and post dispatching this call.
 		#[pallet::call_index(4)]
 		#[pallet::weight({
-		let dispatch_info = call.get_dispatch_info();
-		(dispatch_info.weight + < T as Config >::WeightInfo::pre_sponsor() + < T as Config >::WeightInfo::post_sponsor() + T::DbWeight::get().reads_writes(2, 2), dispatch_info.class, Pays::No)
-		})]
+		let dispatch_infos = calls.iter().map(|call| call.get_dispatch_info()).collect::<Vec<_>>();
```
simonsso, line 435: Compare to https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/utility/src/lib.rs#L305C1-L322C6

pallets/sponsorship/src/tests.rs
```diff
@@ -2519,3 +2519,303 @@ fn sponsor_call_is_invalid_if_sponsor_account_is_running_low() {
...
+		let user = 2u64;
+
+		assert_ok!(SponsorshipModule::register_users(
+			RuntimeOrigin::signed(pot_details.sponsor),
+			pot,
+			vec![user],
+			user_fee_quota,
+			user_reserve_quota
+		));
+		let _user_details = User::<Test>::get(pot, user).unwrap();
```
aliXsed, line 2790: You can remove this line just to make the test focused on its purpose. `register_users` is tested in our other unit tests and if it's ok we can proceed.

pallets/sponsorship/src/tests.rs
```diff
@@ -2519,3 +2519,303 @@ fn sponsor_call_is_invalid_if_sponsor_account_is_running_low() {
...
+		}));
+
+		let _pre_dispatch_details = ChargeSponsor::<Test>::default()
+			.pre_dispatch(
+				&user,
+				&sponsor_for_uniques_create_call,
+				&sponsor_for_uniques_create_call.get_dispatch_info(),
+				0,
+			)
+			.ok();
```
aliXsed, line 2814: Again removing these lines helps with the focus on the purpose of this test.

pallets/sponsorship/src/tests.rs
```diff
@@ -2519,3 +2519,303 @@ fn sponsor_call_is_invalid_if_sponsor_account_is_running_low() {
...
+		);
+		assert_eq!(
+			Balances::total_balance(&user_details.proxy),
+			user_details_post_dispatch.reserve_quota.balance()
+		);
+	});
+}
+
+#[test]
+fn compound_transaction_illegal_content_should_be_filtered() {
```
aliXsed, line 2754: The test title doesn't define what is illegal and I couldn't notice it easily by reading the test either.

pallets/sponsorship/src/tests.rs
```diff
@@ -2519,3 +2519,303 @@ fn sponsor_call_is_invalid_if_sponsor_account_is_running_low() {
...
+
+		let pre_dispatch_details = ChargeSponsor::<Test>::default()
+			.pre_dispatch(
+				&user,
+				&sponsor_for_uniques_create_call,
+				&sponsor_for_uniques_create_call.get_dispatch_info(),
+				0,
+			)
+			.ok();
+		assert!(SponsorshipModule::sponsor_for(RuntimeOrigin::signed(user), pot, unique_create_call.clone()).is_err());
```
aliXsed, line 2709: `asser_noop` is a lot more expressive than assert!(...is_err())

pallets/sponsorship/src/tests.rs
```diff
@@ -2519,3 +2519,303 @@ fn sponsor_call_is_invalid_if_sponsor_account_is_running_low() {
...
+		assert_ok!(SponsorshipModule::register_users(
+			RuntimeOrigin::signed(pot_details.sponsor),
+			pot,
+			vec![user],
+			user_fee_quota,
+			user_reserve_quota
+		));
+		let user_details = User::<Test>::get(pot, user).unwrap();
+
+		let unique_create_call = vec![
```
aliXsed, line 2677:
```suggestion
		let batched_uniques_and_non_uniques_calls = vec![
```

pallets/sponsorship/src/tests.rs
```diff
@@ -2519,3 +2519,303 @@ fn sponsor_call_is_invalid_if_sponsor_account_is_running_low() {
...
+			Balances::free_balance(pot_details.sponsor),
+			pot_reserve_quota
+				- fee - user_details_post_dispatch.reserve_quota.balance()
+				- PotDeposit::get()
+				- UserDeposit::get()
+		);
+		assert_eq!(
+			Balances::total_balance(&user_details.proxy),
+			user_details_post_dispatch.reserve_quota.balance()
+		);
```
aliXsed, line 2749: This whole part can be removed from this test because they are testing things not related to the main purpose of this test.

pallets/sponsorship/src/tests.rs
```diff
@@ -2519,3 +2519,303 @@ fn sponsor_call_is_invalid_if_sponsor_account_is_running_low() {
...
+
+		let unique_create_call = vec![
+			Box::new(RuntimeCall::Uniques(pallet_uniques::Call::create {
+				collection: 0u32,
+				admin: user,
+			})),
+			Box::new(RuntimeCall::Uniques(pallet_uniques::Call::create {
+				collection: 1u32,
+				admin: user,
+			})),
```
aliXsed, line 2570: Can you instead add another uniques call here which relies on the successful execution of the previous one? For example can you mint an item for that collection instead?

aliXsed:
@simonsso it's a nice clear approach. I like it. I have just added some comments to improve the tests.

aliXsed (approved):

PR #837, opened 2024-03-21
Upgrade gh action versions

simonsso:


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/.github/workflows/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
```
aliXsed, line 10: Can we go monthly? Any chain changes related to the dependencies is not quite straightforward to address without proper upgrades.

simonsso: It is for the CI actions on the chain. I don't expect this one to be noisy, but I will revert to monthly if it gets out of hand.

PR #843, opened 2024-03-26
Add workflow for paradis try-runtime

simonsso:


.github/workflows/docker.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
   container_gpr:
     name: Build and push docker image
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-8-core
```
simonsso, line 15:
```suggestion
    runs-on: ubuntu-latest
```

Unchange this

PR #854, opened 2024-04-12
Iso/20240412 t0953 docker

simonsso:


simonsso (commented):
No speedup I was expecting but the container code is at least simplified.

PR #624, opened 2022-06-17
fix(pallet-grants): prevent a transaction failure

aliXsed:
prevent a transaction failure when cancel_all_vesting_schedules was not limited to the free balance

PR #872, opened 2024-07-16
feat(runtime-eden): add bridge_all_vesting_schedules

aliXsed:
1. The bridge is initiated by the user/grant owner similar to how bridging of free tokens currently happening.
2. The bridge_all_vesting_schedules function is added to to pallet-grants which will be very similar to cancel_all_vesting_schedules with the difference that for this one only the owner of the grant and not the governance of the chain can initiate a bridge for locked tokens.
3. Similar to cancel_… bridge_all_vesting_schedules  will first do a claim for the user so they get all their free-able tokens into their own account first.
4. bridge_all_vesting_schedules burns the locked grants of the user and remove their details as well if it succeeds.
5. bridge_all_vesting_schedules will then generate an event traceable by our bridge off-chain oracles which indicates a bridge operation is complete on the parachain side with the user’s given zksync address as the destination.

- [x] Add benchmark test
- [x] Do an initial benchmark

### DIFF
--- a/pallets/grants/src/lib.rs
+++ b/pallets/grants/src/lib.rs
@@ -215,6 +215,8 @@ pub mod pallet {
 			Ok(().into())
 		}
 
+		/// Initiate a bridge transfer of all vested funds to the given `eth_address` on the given `chain_id`.
+		/// This process will need to be completed by the bridge oracles.
 		#[pallet::call_index(4)]
 		#[pallet::weight(T::WeightInfo::bridge_all_vesting_schedules())]
 		pub fn bridge_all_vesting_schedules(
@@ -225,6 +227,12 @@ pub mod pallet {
 			let from = ensure_signed(origin)?;
 
 			let locked_amount_left = Self::do_claim(&from);
+			if locked_amount_left.is_zero() {
+				<VestingSchedules<T>>::remove(from);
+				Self::deposit_event(Event::NoVestedFundsToBridgeAfterClaim);
+				return Ok(().into());
+			}
+
 			let free_balance = T::Currency::free_balance(&from);
 			let bridgeable_funds = locked_amount_left.min(free_balance);
 
@@ -238,9 +246,8 @@ pub mod pallet {
 			.map_err(|_| Error::<T>::FailedToSettleBridge)?;
 
 			let grants = <VestingSchedules<T>>::take(&from);
-			<VestingSchedules<T>>::remove(from);
 
-			Self::deposit_event(Event::Bridge {
+			Self::deposit_event(Event::BridgeInitiated {
 				to: eth_address,
 				chain_id,
 				amount: bridgeable_funds,
@@ -262,12 +269,17 @@ pub mod pallet {
 		VestingSchedulesCanceled(T::AccountId),
 		/// Renounced rights to cancel grant for the given account id \[who\]
 		Renounced(T::AccountId),
-		Bridge {
+		/// Initiated a bridge transfer of all vested funds \[to, chain_id, amount, grants\]
+		/// The field amount is crucial for the bridge because it shows the total entitlement of the user on the other side of the bridge.
+		BridgeInitiated {
 			to: [u8; 20],
 			chain_id: u64,
 			amount: BalanceOf<T>,
 			grants: BoundedVec<VestingScheduleOf<T>, T::MaxSchedule>,
 		},
+		/// Bridge was initiated and successfully completed one sidedly because
+		/// there were no vested funds to bridge after claiming free tokens for the user.
+		NoVestedFundsToBridgeAfterClaim,
 	}
 
 	#[pallet::error]
--- a/pallets/grants/src/lib.rs
+++ b/pallets/grants/src/lib.rs
@@ -243,7 +243,6 @@ pub mod pallet {
 				Error::<T>::BridgeNotFound
 			);
 
-			ensure!(!Self::renounced(from.clone()), Error::<T>::Renounced);
 			let locked_amount_left = Self::do_claim(&from);
 			if locked_amount_left.is_zero() {
 				<VestingSchedules<T>>::remove(from);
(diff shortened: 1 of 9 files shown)


pallets/grants/src/lib.rs
```diff
@@ -214,6 +214,48 @@ pub mod pallet {
...
+
+			let locked_amount_left = Self::do_claim(&from);
+			if locked_amount_left.is_zero() {
+				<VestingSchedules<T>>::remove(from);
+				Self::deposit_event(Event::NoVestedFundsToBridgeAfterClaim);
+				return Ok(().into());
+			}
+
+			let free_balance = T::Currency::free_balance(&from);
+			let bridgeable_funds = locked_amount_left.min(free_balance);
```
ETeissonniere, line 237: I am not sure why this is necessary, can you clarify? I assume you are trying to cover some edge case?

pallets/grants/src/lib.rs
```diff
@@ -214,6 +214,48 @@ pub mod pallet {
...
 		}
+
+		/// Initiate a bridge transfer of all vested funds to the given `eth_address` on the given `chain_id`.
+		/// This process will need to be completed by the bridge oracles.
+		#[pallet::call_index(4)]
+		#[pallet::weight(T::WeightInfo::bridge_all_vesting_schedules())]
+		pub fn bridge_all_vesting_schedules(
+			origin: OriginFor<T>,
+			eth_address: [u8; 20],
+			chain_id: u64,
```
ETeissonniere, line 225: Shall we have something similar to the ink! contract where we verify the address' checksum & chain id instead of adding it to the event at the bottom?

pallets/grants/src/lib.rs
aliXsed: It's exactly equal to cancel_all have a look at that. Yes it's covering a corner case

pallets/grants/src/lib.rs
aliXsed: We would need to have it in the event anyway. But I agree if we can verify it we should do it as well. I will add it soon

pallets/grants/src/lib.rs
aliXsed: Let's have a look at https://github.com/NodleCode/chain/pull/624

aliXsed: It's a historical thing. In the past some of the vesting calls where not safe and they could cause corrupt state. In those corrupt state, the user could end up having less free coins that was granted to them. This `minimum` check ensures the call will succeed safely even for those failures.

pallets/grants/src/lib.rs
aliXsed: Actually the verify became sort of useless on the side of the smart contract for 2 reasons:
1. Our own Nodle Client and App convert any address to a valid address as recognized by the smart contract.
2. ZkSync has not opted https://eips.ethereum.org/EIPS/eip-1191

pallets/grants/src/lib.rs
```diff
@@ -214,6 +227,91 @@ pub mod pallet {
...
+			eth_address: [u8; 20],
+			bridge_id: u32,
+		) -> DispatchResultWithPostInfo {
+			let from = ensure_signed(origin)?;
+			ensure!(
+				Bridges::<T>::contains_key(&BridgeId(bridge_id)),
+				Error::<T>::BridgeNotFound
+			);
+
+			ensure!(!Self::renounced(from.clone()), Error::<T>::Renounced);
```
ETeissonniere, line 246: I do not think this is necessary here since we do not really cancel the vesting schedules

pallets/grants/src/lib.rs
```diff
@@ -214,6 +227,91 @@ pub mod pallet {
...
+			Ok(().into())
+		}
+
+		/// Allow governance to indicate a bridge is setup between the parachain and a remote chain.
+		#[pallet::call_index(5)]
+		#[pallet::weight(T::WeightInfo::set_bridge())]
+		pub fn set_bridge(
+			origin: OriginFor<T>,
+			bridge_id: u32,
+			bridge_name: Vec<u8>,
```
ETeissonniere, line 284: Why is `bridge_name` necessary? Just UX?
We could also simplify towards chain ID being the universal ID to manage bridges

pallets/grants/src/tests.rs
```diff
@@ -285,6 +286,282 @@ fn cancel_auto_claim_recipient_funds_and_wire_the_rest() {
 	});
 }
```
ETeissonniere, line 288: I might be pushing it, but it might be a good thing to get tests for the `ensure_root` statements as to ensure they do not get removed by mistake. Meaning two tests: `remove_bridge_ensures_root` and `set_bridge_ensures_root` (or whichever name is best).

pallets/grants/src/lib.rs
aliXsed: 1. bridge_name will improve the UX a lot but only occupies 32 bytes.
2. Can you explain your second point more?

pallets/grants/src/tests.rs
aliXsed: Good suggestion. I will be adding them.

pallets/grants/src/lib.rs
aliXsed: That's a very good catch. Thank you for spotting it. Copy paste issue

aliXsed: I fixed it after adding a test that could catch this issue

pallets/grants/src/tests.rs
aliXsed: Done

ETeissonniere (approved):

pallets/grants/src/lib.rs
simonsso: Vested schedule not canceled -> Tokens claimed -> Token transferred to new account -> token briged ?

PR #875, opened 2024-08-05
Upgrade to polkadot-v1.13.0

aliXsed:
Among the changes done for this implementation we have the StorageWeightReclaim feature which is introduced in [Polkadot v1.9.0](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.9.0) and no issues reported about it since then. However my investigation shows the polkadot system parachains have not adopted it yet. This feature would allow the differences between benchmarked proof sizes and the actual proof sizes to be accounted for plus a facility for manual weight metering inside on_idle. This needs both the client and the runtime to support it. I noticed we had brought a part of this feature to our runtime. In this PR our client and runtime will become compatible about this check. My own assessment is this should be a pretty safe adoption especially if no problems are discovered during testing and try-runtime. The good thing about keeping it is that our upgrade will become closer to the upstream template/suggestion for parachains. 
- [x] Remove xToken 
- [x] Update Dependencies
- [x] Fix XCM build while sticking to v3
- [x] Fix Runtime
- [x] Fix Node (Update jsonrpsee here)
- [x] Fix tests
- [x] Fix benchmarks
- [x] Fix and test try-runtime
- [x] Bump up all our crate versions
- [x] Fix `on-chain-release-build` feature for eden-runtime
- [x] Update CI to produce the runtime with `on-chain-release-build` feature
- [x] Add and enable the rest of the benchmark command
- [x] Bake eden-dev chainspec into runtime which is useful for benchmarking
- [x] Remove redundant code in chain-spec
- [x] Verify new frame_system::Config
- [x] Check if async backing is properly supported
- [x] Verify block duration for parachain (6s according to polkadot parachain templates or 12s as before and according to parachain-commons
- [x] Verify the necessity of `StorageWeightReclaim`: https://github.com/paritytech/polkadot-sdk/pull/3002
- [x] Review XCM config
- [x] Check if we need to 	
#[cfg(feature = "runtime-benchmarks")]
	type XcmpQueue = ();  => Result: it doesn't seem so and it's not fixing xcm benchmarking
- [x] Fix collator-selection benchmarking
- [x] Fix membership benchmarking
- [x] Fix fungible benchmarking
- [x] Fix contracts benchmarking
- [x] Fix XCM benchmarking 
- [x] Fix generic benchmarking
- [x] devcontainer CI build failing
- [x] Add any migration logic needed
- [x] Fix front-run risk in cumulus_pallet_xcmp_queue::migration::v5
- [x] Test XCM transfer with Zombienet
- [x] Do benchmark and update weights


runtimes/eden/src/migration.rs
```diff
@@ -0,0 +1,66 @@
...
+		let pre_image_status_for_key = [
+			216_u8, 243, 20, 183, 244, 230, 176, 149, 240, 248, 238, 70, 86, 164, 72, 37, 85, 177, 174, 142, 206, 213,
+			82, 47, 60, 64, 73, 188, 132, 237, 164, 168, 134, 165, 126, 209, 14, 33, 121, 241, 141, 181, 169, 30, 183,
+			253, 57, 95, 170, 112, 42, 188, 5, 207, 200, 241, 243, 215, 36, 138, 208, 108, 237, 135,
+		];
+		let _ = frame_support::storage::unhashed::clear_prefix(&pre_image_status_for_key, Some(1), None);
+
+		StorageVersion::new(1).put::<pallet_uniques::Pallet<T>>();
+
+		T::DbWeight::get().writes(3)
```
ETeissonniere, line 46: What is this doing and is it necessary or a relicate from a prior migration?

aliXsed: No it's actually necessary. The try-runtime test helps us spot un-decodable memory storage items which are indications of missed migrations in the past. We found them when we were preparing the previous upgrade which was deployed on the testnet but didn't end up on mainnet. We did are research about these specific storage items and the safest way to handle them is to remove them.

ETeissonniere (approved):

